Issue #4, opened 2020-03-02
Integrate with Transit Gateway


dav3r:
Integrate the assessment VPC with the Transit Gateway in our Shared Services account.

dav3r:
I think it's safe to close out this issue now that we have confirmed that traffic is successfully being routed between the Shared Services account VPC and the `env0` account Assessment VPC (see #5).  Any further Transit Gateway integration work will be tracked in new issues.

Issue #2, opened 2020-03-02
Add Guacamole Instance to VPC


felddy:
@dav3r commented on [Mon Mar 02 2020](https://github.com/cisagov/cool-system/issues/8)

Add Terraform to spin up a Guacamole instance in the VPC.

PR #5, opened 2020-03-10
Create Assessment VPC with Guacamole Instance

### LINKED ISSUE #1: Create VPC
Write the Terraform for the basic assessment VPC.

dav3r:
# 

## 🗣 Description
This is a ~somewhat~ large PR, but it moves the assessment environment from being completely empty to being in a good place for us to add on to.  Here's what is defined by the Terraform in this PR:
- A VPC with one "Operations" (public) subnet and one or more private subnets (number of private subnets can be specified in the `tfvars` file)
- An Internet Gateway for the Operations subnet
- A NAT Gateway for each private subnet
- A connection to the Transit Gateway in the Shared Services account (untested as of now)
- Routing tables for each subnet, including routes to the Transit Gateway (again, untested)
- Network ACLs for the public and private subnets, with a reasonable set of rules (they will almost certainly need changes later, but this should be a good start)
- Two security groups: 
  - "Desktop Gateway" (for the Guacamole instance), with a reasonable initial set of rules
  - "Operations" (for the operating instances, like Kali, etc.) - no rules have been set up yet in this SG
- A Guacamole instance in one of the private subnets - note that there are no other instances yet for Guacamole to connect to, but we'll get there in a PR in the very near future.

Note that I am using a bunch of remote Terraform states in order to minimize the number of required input variables.  I am aware that we may consolidate our remote state data in the future- if/when that happens, some changes may be needed here.

Resolves #1.  Starts the work of #2 and #4, but there is still more to be done on them.

## 💭 Motivation and Context
We need to get `env0` up and running for the start of user acceptance testing.  This PR lays down a good chunk of Terraform that we need and can build on.

Next steps for future PRs:
- Add one Kali instance (and support having additional instances)
- Add DNS records so the Guacamole instance can be reached via `guac.env0.cool.cyber.dhs.gov` (for example)
- Add private DNS records so that Operations instances can be reached via `kali0.env0` (for example)
- Update `cloud-init` scripts so that the correctly set up connections (within Guacamole) to all instances in the Operations subnet
- Update the README with a bit more useful info 😜 

## 🧪 Testing
I successfully applied all of this Terraform in the `env0` account.  Since the OpenVPN networking is still in flux, I ssh'd to the OpenVPN instance and used that as a bastion so that I could ssh to the `env0` Guacamole instance.  Once I was on the Guacamole instance, I confirmed that the Docker composition (Guacamole / Postgres / nginx) was running as expected.

Next, I opened up an ssh tunnel (again via the OpenVPN instance) and forwarded port 8443 on my local workstation to port 8443 on the Guacamole instance, where nginx is listening.  That allowed me to hit `https://localhost:8443` in a browser on my workstation and I was able to successfully operate the Guacamole GUI from the Guacamole instance in `env0`.  I couldn't do a whole lot since there are no other instances to connect to via Guacamole, but at least I knew Guacamole was up and running.

## 📷 Screenshots

Screen Shot 2020-03-09 at 11 23 00 PM
Screen Shot 2020-03-09 at 11 25 04 PM
Screen Shot 2020-03-09 at 11 25 22 PM
Screen Shot 2020-03-09 at 11 25 35 PM

## 🚥 Types of Changes

- [x] New feature (non-breaking change which adds functionality)

## ✅ Checklist


dav3r:
Note that "TBD" throughout this PR is my placeholder for the first Kali instance in the Operations subnet, which will be coming soon.

desktop_gateway_sg_rules.tf
```diff
@@ -0,0 +1,94 @@
+# Allow ingress from COOL Shared Services via ssh
+# For: DevOps ssh access to Guacamole instance
+resource "aws_security_group_rule" "desktop_gw_ingress_from_cool_via_ssh" {
+  provider = aws.provisionassessment
+
+  security_group_id = aws_security_group.desktop_gateway.id
+  type              = "ingress"
+  protocol          = "tcp"
+  cidr_blocks       = [local.cool_shared_services_cidr_block]
```
jsf9k: We should consider limiting this to the VPN client CIDR.  You could add the VPN instance IP as a second block for now while you are using it as a jump box.

desktop_gateway_sg_rules.tf
```diff
@@ -0,0 +1,94 @@
...
+  # ipv6_cidr_blocks  = TBD
+  from_port = 22
+  to_port   = 22
+}
+
+# # Allow egress via ssh to the assessment operating instance(s)
```
jsf9k: You could probably allow ssh egress to the entire ENV* VPC CIDR, or at least the operations subnet.

desktop_gateway_sg_rules.tf
```diff
@@ -0,0 +1,94 @@
...
+
+  security_group_id = aws_security_group.desktop_gateway.id
+  type              = "egress"
+  protocol          = "tcp"
+  cidr_blocks       = ["0.0.0.0/0"]
+  from_port         = 443
+  to_port           = 443
+}
+
+# Allow ingress from COOL Shared Services via port 8443 (nginx/guacamole web)
```
jsf9k: Again, this should be limited to the VPN client CIDR.

desktop_gateway_sg_rules.tf
```diff
@@ -0,0 +1,94 @@
...
+  security_group_id = aws_security_group.desktop_gateway.id
+  type              = "ingress"
+  protocol          = "tcp"
+  cidr_blocks       = [local.cool_shared_services_cidr_block]
+  # ipv6_cidr_blocks  = TBD
+  from_port = 8443
+  to_port   = 8443
+}
+
+# # Allow egress via VNC to the assessment operating instance(s)
```
jsf9k: You could allow egress to the entire ENV* VPC here instead of waiting until you can specify the instance IPs.

felddy (approved):
Good job using TBDs..  I appreciate the comments about the template tool, and the IPv6.

😎

guacamole_ec2.tf
```diff
@@ -0,0 +1,50 @@
...
+  most_recent = true
+}
+
+# The Guacamole EC2 instance
+resource "aws_instance" "guacamole" {
+  provider = aws.provisionassessment
+
+  ami                  = data.aws_ami.guacamole.id
+  availability_zone    = "${var.aws_region}${var.aws_availability_zone}"
+  iam_instance_profile = aws_iam_instance_profile.guacamole.name
```
jsf9k, line 32: Eventually you're going to need to add the CloudwatchAgentServerPolicy to this instance profile.  See [here](https://github.com/cisagov/freeipa-master-tf-module/blob/develop/instance_role.tf#L42-L46) for more details.

network_acls.tf
```diff
@@ -0,0 +1,40 @@
...
+
+  tags = merge(
+    var.tags,
+    {
+      "Name" = "Assessment Operations"
+    },
+  )
+}
+
+# ACLs for the private subnet of the VPC
```
jsf9k: "subnet" --> "subnets"

private_acl_rules.tf
```diff
@@ -0,0 +1,113 @@
...
+resource "aws_network_acl_rule" "private_ingress_from_cool_via_ssh" {
+  provider = aws.provisionassessment
+  for_each = toset(var.private_subnet_cidr_blocks)
+
+  network_acl_id = aws_network_acl.private[each.value].id
+  egress         = false
+  protocol       = "tcp"
+  rule_number    = 100 + index(var.private_subnet_cidr_blocks, each.value)
+  rule_action    = "allow"
+  cidr_block     = local.cool_shared_services_cidr_block
```
jsf9k: Again, this should be restricted to the VPN client CIDR.

vnc_assume_role_policy_doc.tf
```diff
@@ -0,0 +1,32 @@
+# ------------------------------------------------------------------------------
+# Create an IAM policy document that allows the EC2 instances in the
```
jsf9k: This policy document does more than what is described in this comment.

subnets.tf
```diff
@@ -0,0 +1,70 @@
...
+}
+
+resource "aws_nat_gateway" "nat_gws" {
+  provider = aws.provisionassessment
+
+  for_each = toset(var.private_subnet_cidr_blocks)
+
+  allocation_id = aws_eip.nat_gw_eips[each.value].id
+  # Reminder: NAT gateways live in the operations subnet
+  subnet_id = aws_subnet.operations.id
```
jsf9k, line 64: It doesn't make sense to use multiple NAT GWs if they are all in the same subnet, and hence in the same AZ.  The point of multiple NAT GWs is to allow some operations to continue if an AZ goes down.  You may as well just use a single NAT GW in the operations subnet.

What I did in cisagov/cool-sharedservices-networking was create three private subnets in three different AZs and three public subnets in the same three AZs.  Each private subnet uses a NAT GW located in the public subnet that lives in the same AZ.  You could also do that here, which would be even better.  That way operations could continue even if an AZ goes down.  They would just have to switch to a different operations subnet located in a different AZ.

jsf9k (changes requested):
I'm not sure that changes are necessary, but I brought up several points in my comments that I'd like to see either negated or made into issues before approving this PR.

network_acls.tf
dav3r: Fixed in https://github.com/cisagov/cool-assessment-terraform/pull/5/commits/00767e0675736ea48ecfb78ea2f17e6ba9670feb.

vnc_assume_role_policy_doc.tf
dav3r: @jsf9k Does this wording sound better to you?
```
# Create an IAM policy document that allows EC2 instances and any user in the
# assessment account to assume the role this policy is attached to.
```

subnets.tf
dav3r: Excellent catch @jsf9k.  For now, I switched to using a single NAT gateway for all of the private subnets - see https://github.com/cisagov/cool-assessment-terraform/pull/5/commits/b659ddca5fdd40acd596bafadcef402d564400db.

For the future, I created https://github.com/cisagov/cool-assessment-terraform/issues/6.

desktop_gateway_sg_rules.tf
dav3r: Excellent suggestion- I allowed egress to the Operations subnet in https://github.com/cisagov/cool-assessment-terraform/pull/5/commits/18a370b268fe09870bdace5ca7ed24a902a239fd.  This means one less moving part later when we start dynamically deploying instances to the Operations subnet.

desktop_gateway_sg_rules.tf
dav3r: As mentioned above, I allowed egress to the Operations subnet in https://github.com/cisagov/cool-assessment-terraform/pull/5/commits/18a370b268fe09870bdace5ca7ed24a902a239fd.  This means one less moving part later when we start dynamically deploying instances to the Operations subnet.

desktop_gateway_sg_rules.tf
dav3r: See https://github.com/cisagov/cool-assessment-terraform/pull/5/commits/04a8e1f26a2552aed3dbd32fa31549a9b2320c06.

desktop_gateway_sg_rules.tf
dav3r: See https://github.com/cisagov/cool-assessment-terraform/pull/5/commits/04a8e1f26a2552aed3dbd32fa31549a9b2320c06.

private_acl_rules.tf
dav3r: See https://github.com/cisagov/cool-assessment-terraform/pull/5/commits/04a8e1f26a2552aed3dbd32fa31549a9b2320c06.

desktop_gateway_sg_rules.tf
jsf9k: I don't think this is correct with the NAT setup.  Don't you need to add `<openvpn_server_ip>/32` to the list of CIDR blocks?

This reminds me that I have to get OpenVPN + PIV working in Linux as well.

subnets.tf
jsf9k: :+1:

vnc_assume_role_policy_doc.tf
jsf9k: Why is it a user in the assessment account and not the users account?

dav3r: Arg, you are right- there are no users in the assessment account.  

I don't think we even need users to be able to assume this role- it's really just so the cloud-init script on the Guacamole instance can talk to the SSM Parameter Store.  So it will just become:
```
# Create an IAM policy document that allows EC2 instances 
# to assume the role this policy is attached to.
```
And I will remove the second statement in the policy entirely.

dav3r: Cleaned up in https://github.com/cisagov/cool-assessment-terraform/pull/5/commits/f022950aadf7c6b11074df026198c284cc59fe49.

guacamole_ec2.tf
jsf9k: FWIW, no harm is done by adding it now.

desktop_gateway_sg_rules.tf
dav3r: Yup- I was thinking that the OpenVPN server IP was included in the `client_vpn_cidr_block`.  I will change this to `vpn_server_cidr_block` instead.  As @felddy pointed out in a chat, this is preferred over specifying only the IP address of the OpenVPN server because that IP will change whenever the instance is redeployed.  We don't want to have to reapply this terraform in every assessment environment when that happens.

This solution also covers us if multiple VPN servers are deployed for failover situations (as long as they are all within `vpn_server_cidr_block`).

dav3r: Take a look at https://github.com/cisagov/cool-assessment-terraform/pull/5/commits/dcd6a4c16ded570a9dea4da64dc55ab038e230a8 and let me know if this looks better.

desktop_gateway_sg_rules.tf
dav3r: Corrected in https://github.com/cisagov/cool-assessment-terraform/commit/dcd6a4c16ded570a9dea4da64dc55ab038e230a8.

private_acl_rules.tf
dav3r: Corrected in https://github.com/cisagov/cool-assessment-terraform/commit/dcd6a4c16ded570a9dea4da64dc55ab038e230a8.

guacamole_ec2.tf
dav3r: No time like the present!  See https://github.com/cisagov/cool-assessment-terraform/pull/5/commits/0677acd8b8ec36508c2bb6e58c60efa89956025a.

jsf9k (approved):

vnc_assume_role_policy_doc.tf
dav3r: @jsf9k - After I deployed this PR, I noticed that one of the cloud-init scripts was failing with this error:
```
botocore.exceptions.ClientError: An error occurred (AccessDenied)
when calling the AssumeRole operation:
User: arn:aws:sts::051445540311:assumed-role/guacamole_instance_role_default/i-059a2b3e51732a452
is not authorized to perform: sts:AssumeRole on resource:
arn:aws:iam::207871073513:role/ParameterStoreReadOnly-env0-VNC
```
It seems that the statement I removed in https://github.com/cisagov/cool-assessment-terraform/commit/f022950aadf7c6b11074df026198c284cc59fe49 was allowing the Guacamole instance to successfully assume the `ParameterStoreReadOnly-env0-VNC` role.  I'm pretty sure that statement was allowing any user from the `env0` account to assume the role and that apparently includes the Guacamole instance profile.

I have tested an updated policy that only allows the Guacamole instance profile role to assume the `ParameterStoreReadOnly-env0-VNC` role, so I will be creating a PR with that change soon.

PR #8, opened 2020-03-11
Fix trust relationship for ParameterStoreReadOnly-VNC role

dav3r:
# 

## 🗣 Description
This PR correctly sets the Guacamole instance profile role to be the only trusted entity (so far) that is allowed to assume the ParameterStoreReadOnly-VNC role.

## 💭 Motivation and Context
As mentioned [here](https://github.com/cisagov/cool-assessment-terraform/pull/5#discussion_r390723444), after I deployed https://github.com/cisagov/cool-assessment-terraform/pull/5, I noticed that one of the cloud-init scripts on the Guacamole instance was failing with this error:
```
botocore.exceptions.ClientError: An error occurred (AccessDenied)
when calling the AssumeRole operation:
User: arn:aws:sts::051445540311:assumed-role/guacamole_instance_role_default/i-059a2b3e51732a452
is not authorized to perform: sts:AssumeRole on resource:
arn:aws:iam::207871073513:role/ParameterStoreReadOnly-env0-VNC
```
It seems that the statement I removed in https://github.com/cisagov/cool-assessment-terraform/commit/f022950aadf7c6b11074df026198c284cc59fe49 was allowing the Guacamole instance to successfully assume the `ParameterStoreReadOnly-env0-VNC` role.  I'm pretty sure that statement was allowing any user from the `env0` account to assume the role and that apparently included the Guacamole instance profile.

## 🧪 Testing
I applied the updated policy in this PR and verified that the newly-deployed Guacamole instance was able to assume the role in question and successfully run the cloud-init script that was previously failing.

## 🚥 Types of Changes

## ✅ Checklist

### DIFF
--- a/vnc_assume_role_policy_doc.tf
+++ b/vnc_assume_role_policy_doc.tf
@@ -1,6 +1,6 @@
 # ------------------------------------------------------------------------------
-# Create an IAM policy document that allows EC2 instances
-# to assume the role this policy is attached to.
+# Create an IAM policy document that only allows specific instance profile
+# roles to assume the role this policy is attached to.
 # ------------------------------------------------------------------------------
 
 data "aws_iam_policy_document" "vnc_assume_role_doc" {
@@ -10,9 +10,9 @@ data "aws_iam_policy_document" "vnc_assume_role_doc" {
     ]
 
     principals {
-      type = "Service"
+      type = "AWS"
       identifiers = [
-        "ec2.amazonaws.com",
+        aws_iam_role.guacamole_instance_role.arn
       ]
     }
   }


jsf9k (approved):
Interesting.  I really thought the service would be sufficient here.

dav3r:
> Interesting. I really thought the service would be sufficient here.

My guess is that may be true in a single account setup, but for a cross-account setup like ours, I think we need to be more specific about who is trusted to assume the role.

[This doc](https://docs.amazonaws.cn/en_us/IAM/latest/UserGuide/id_roles_use_switch-role-ec2.html#roles-usingrole-ec2instance-permissions) was helpful in clarifying things for me.

felddy (approved):
Makes sense.  Looks good.  Very hansom PR.  :shipit:

PR #14, opened 2020-03-16
Fix incorrect SG rule allowing access to port 5901 (VNC)

dav3r:
# 

## 🗣 Description
This PR changes the security group rules in the Operations SG so that port 5901 is only accessible from the Guacamole instance.

## 💭 Motivation and Context
This was the original intent of this security group rule, however it was accidentally created incorrectly, allowing access to port 5901 from anywhere.  Since the corresponding network ACL was set up correctly (deny public access to port 5901), the only way this could've been taken advantage of was from within the Operations security group...

...And that is exactly how it was exploited, as @KyleEvers helpfully pointed out that the Kali instances could VNC to one another.  Big thanks to @KyleEvers for noticing this so quickly and allowing us to fix it!

## 🧪 Testing
From the `kali0.env0` instance, I verified that I was able to VNC to `kali1.env0`.  Then I applied the Terraform changes in this PR and confirmed that I was no longer able to VNC from `kali0` to `kali1`.

## 🚥 Types of Changes

## ✅ Checklist

### DIFF
--- a/operations_sg_rules.tf
+++ b/operations_sg_rules.tf
@@ -72,7 +72,7 @@ resource "aws_security_group_rule" "operations_ingress_from_anywhere_via_ports_3
 # Allow ingress from anywhere via ephemeral TCP/UDP ports above 5901 (5902-65535)
 # For: Assessment team operational use, but don't want to allow
 #      public access to VNC on port 5901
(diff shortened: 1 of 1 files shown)


jsf9k (approved):
Good catch!

mcdonnnj (approved):
image

felddy (approved):
Nice.  Good find @KyleEvers 
💯

PR #21, opened 2020-03-19
Stop hardcoding workspace name in remote state configurations

jsf9k:
## 🗣 Description

The workspace name in remote state configurations was previously hardcoded to "production".  This PR changes the code to instead use `terraform.workspace`, so that the workspace name changes as we switch between the production and staging environments.

## 💭 Motivation and Context

We need to be able to switch as seamlessly as possible between the production and staging environments.  The more manual foo that is required the more likely that we make a mistake and damage production.

## 🧪 Testing

These changes were successfully deployed to staging.

## 🚥 Types of Changes

## ✅ Checklist

### DIFF
--- a/remote_states.tf
+++ b/remote_states.tf
@@ -44,6 +44,9 @@ data "terraform_remote_state" "dynamic_assessment" {
     key            = "cool-accounts/dynamic.tfstate"
   }
 
+  # Note that this workspace is different from all the others.  For
+  # the others we want production, staging, etc.  Here, though, we
+  # want (for example) env0, (for production), env0-staging, etc.
   workspace = local.workspace_name
 }
 
--- a/remote_states.tf
+++ b/remote_states.tf
@@ -15,7 +15,7 @@ data "terraform_remote_state" "dns_certboto" {
     key            = "cool-dns-certboto/terraform.tfstate"
   }
 
-  workspace = "production"
+  workspace = terraform.workspace
 }
 
 data "terraform_remote_state" "dynamic_assessment" {
@@ -45,7 +45,7 @@ data "terraform_remote_state" "images" {
     key            = "cool-accounts/images.tfstate"
   }
 
-  workspace = "production"
+  workspace = terraform.workspace
 }
 
 data "terraform_remote_state" "images_parameterstore" {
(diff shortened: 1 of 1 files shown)


dav3r:
@jsf9k - I don't think this will work, since the workspaces in this project are `env0`, `env1`, etc.

jsf9k:
Something needs to be done with the remote state configurations whose `workspace` attribute is set to `var.assessment_account_name`.  How about `terraform.workspace == "production" ? var.assessment_account_name : "${var.assessment_account_name}-${terraform.workspace}"`?

Ideally we could make the existing workspace names `env0-production`, for example, but that would involve renaming some workspaces.  The TF code above assumes we are unwilling to do that.

mcdonnnj (approved):
This is the good stuff 🍻

jsf9k:
OK, @dav3r, please have another look when you have time.

remote_states.tf
```diff
@@ -45,7 +59,7 @@ data "terraform_remote_state" "images" {
     key            = "cool-accounts/images.tfstate"
   }
 
-  workspace = "production"
+  workspace = local.workspace_type
```
felddy, line 65: Shouldn't this be `local.workspace_name` ?
(turtles all the way down) ⬇️

felddy (approved):
~I think there is a few bugs.  See my comments.~
~Otherwise,~ I think this will work.

remote_states.tf
jsf9k: I don't think so.  `local.workspace_name` would be something like "env0-staging".  `local.workspace_type` would be "staging".  Or are you seeing something I am missing?

felddy: It's all good.  I was misreading the code, and was confused at line 47.

remote_states.tf
```diff
@@ -30,7 +44,7 @@ data "terraform_remote_state" "dynamic_assessment" {
     key            = "cool-accounts/dynamic.tfstate"
   }
 
-  workspace = var.assessment_account_name
+  workspace = local.workspace_name
```
dav3r, line 50: It's probably worth a brief comment here to explain why this one is different from the others (even though things are well-explained in the comment at the top of this file).

dav3r (approved):
Yessir, this will do nicely. 👍 
Please consider my one request for an additional comment.

remote_states.tf
felddy: This is what threw me off.  Until I expanded the whole file.

jsf9k: Good idea.  Comment added in commit f9b0484.

PR #24, opened 2020-03-26
Fix EFS and new Transit Gateway code

jsf9k:
## 🗣 Description

When I tried to deploy to the "env0 (Staging)" account, I ran into several issues with the EFS code and the new Transit Gateway code that uses route tables specific to each assessment environment.

## 💭 Motivation and Context

EFS (persistent storage) and assessment environments that are isolated from one another are both required for the COOL.

## 🧪 Testing

I used these changes to successfully deploy to the "env0 (Staging)" account.

## 🚥 Types of Changes

## ✅ Checklist


locals.tf
```diff
@@ -35,12 +35,32 @@ locals {
...
+  assessment_account_name = [
+    for account in data.aws_organizations_organization.cool.accounts :
+    account.name
+    if account.id == local.assessment_account_id
+  ][0]
+
+  # Determine assessment account type based on account name.
+  #
+  # The account name format is "ACCOUNT_NAME (ACCOUNT_TYPE)" - for
+  # example, "Shared Services (Production)".
```
dav3r: I like how we have been changing this comment in each project to reflect the type of account, so can we change this to:
`# example, "env0 (Production)".`

locals.tf
```diff
@@ -35,12 +35,32 @@ locals {
...
+  assessment_account_type = length(regexall("\\(([^()]*)\\)", local.assessment_account_name)) == 1 ? regex("\\(([^()]*)\\)", local.assessment_account_name)[0] : "Unknown"
+  workspace_type          = lower(local.assessment_account_type)
+
+  # The Terraform workspace name for this assessment
+  assessment_workspace_name = replace(replace(lower(var.assessment_account_name), "/[()]/", ""), " ", "-")
+
+  assessment_account_name_base = split(" ", var.assessment_account_name)[0]
```
dav3r: We should probably add a comment here that mentions the assumption that our assessment account names are a single word followed by the account type, e.g. "env0 (Production)".

locals.tf
```diff
@@ -35,12 +35,32 @@ locals {
 
   cool_shared_services_cidr_block = data.terraform_remote_state.sharedservices_networking.outputs.vpc.cidr_block
 
-  guacamole_fqdn = format("guac.%s.%s", var.assessment_account_name, var.cool_domain)
+  guacamole_fqdn = format("guac.%s.%s", local.assessment_account_name_base, var.cool_domain)
```
dav3r, line 38: This works for Production environments, but for non-Production environments, I think we want it to be: 
`format("guac.%s.%s.%s", local.assessment_account_name_base, local. workspace_type, var.cool_domain)`

dav3r (changes requested):
I have ~~one change that is necessary (unless I'm misunderstanding something) and~~ a couple of comment-related requests.  Take a look and let me know what you think.

locals.tf
jsf9k: In the case of staging, for example, `var.cool_domain` is `staging.cool.cyber.dhs.gov`.  The change you suggested isn't necessary.

dav3r: Yup, I realized this 2 seconds after I posted it and was just about to confirm that.  Thanks for clarifying.

dav3r (approved):

locals.tf
jsf9k: See commit 60e4037.

locals.tf
jsf9k: See commit b1dc956.

jsf9k: See also commit 10cb0a2, which makes the assumption only that there not be a second "(" in the name.

locals.tf
```diff
@@ -55,8 +55,8 @@ locals {
   assessment_workspace_name = replace(replace(lower(var.assessment_account_name), "/[()]/", ""), " ", "-")
 
   # Note that we are assuming that the assessment account name does
-  # not contain a space character.
-  assessment_account_name_base = split(" ", var.assessment_account_name)[0]
+  # not contain a "(" character.
+  assessment_account_name_base = trimspace(split("(", var.assessment_account_name)[0])
```
dav3r, line 59: Good thinking!

felddy (approved):
I'm not sure I understand everything I'm reading... but the stuff I did understand blew my socks off! 🌋 
I think we'll need a walk-through of EFS and TG so we're all edumacated.

PR #31, opened 2020-04-27
Add support for Nessus instances

### LINKED ISSUE #30: Support Nessus instances
Add the ability to include one or more Nessus instances (see [`nessus-packer`](https://github.com/cisagov/nessus-packer)) in an assessment environment.

dav3r:
# 

## 🗣 Description
This PR adds the ability to include one or more Nessus (see [`nessus-packer`](https://github.com/cisagov/nessus-packer) and [`ansible-role-nessus`](https://github.com/cisagov/ansible-role-nessus/)) instances in an assessment environment.

Resolves #30.

## 💭 Motivation and Context
The assessment teams have told us that Nessus is required in the environment.

NOTES:
- Nessus instances are configured via [a `cloud-init` script](https://github.com/cisagov/cool-assessment-terraform/blob/improvement/add_nessus_instance/cloud-init/nessus-setup.sh) that does the following:
    - Registers Nessus with the activation code (provided via [a new Terraform variable](https://github.com/cisagov/cool-assessment-terraform/blob/improvement/add_nessus_instance/variables.tf#L70-74))
    - Creates a Nessus admin user (optional)
    - Fetches the latest Nessus plugins (only if Nessus registration was done)
    - Rebuilds the Nessus plugin database (only if Nessus registration was done)
- Nessus instances are only accessible via the web browser on a Kali instance in the environment.  If the assessment team operators need to be able to SSH to the Nessus instance, that access will have to be added later.

## 🧪 Testing
All pre-commit checks were successful.

My testing primarily focused on ensuring that the [Nessus setup `cloud-init` script](https://github.com/cisagov/cool-assessment-terraform/blob/improvement/add_nessus_instance/cloud-init/nessus-setup.sh) worked as expected and was robust enough to handle the most likely error situations.

I also verified that the Nessus instance was able to successfully scan and detect vulnerabilities on the two Metasploitable instances that were deployed for COOL User Acceptance Testing.

## 🚥 Types of Changes

- [x] New feature (non-breaking change which adds functionality)

## ✅ Checklist

### DIFF
--- a/nessus_ec2.tf
+++ b/nessus_ec2.tf
@@ -1,0 +1,53 @@
+# The Nessus AMI
+data "aws_ami" "nessus" {
+  provider = aws.provisionassessment
+
+  filter {
+    name = "name"
+    values = [
+      "nessus-hvm-*-x86_64-ebs"
+    ]
+  }
+
+  filter {
+    name   = "virtualization-type"
+    values = ["hvm"]
+  }
+
+  filter {
+    name   = "root-device-type"
+    values = ["ebs"]
+  }
+
+  owners      = [local.images_account_id]
+  most_recent = true
+}
+
+# The Nessus EC2 instance
+resource "aws_instance" "nessus" {
+  count    = var.operations_instance_counts["nessus"]
+  provider = aws.provisionassessment
+
+  ami                         = data.aws_ami.nessus.id
+  associate_public_ip_address = true
+  availability_zone           = "${var.aws_region}${var.aws_availability_zone}"
+  iam_instance_profile        = aws_iam_instance_profile.nessus.name
+  instance_type               = "m5.large"
+  subnet_id                   = aws_subnet.operations.id
+
+  root_block_device {
+    volume_type           = "gp2"
+    volume_size           = 128
+    delete_on_termination = true
+  }
+
+  user_data_base64 = data.template_cloudinit_config.nessus_cloud_init_tasks[count.index].rendered
+
+  vpc_security_group_ids = [
+    aws_security_group.efs_client.id,
+    aws_security_group.operations.id,
+  ]
+
+  tags        = merge(var.tags, map("Name", format("Nessus%d", count.index)))
+  volume_tags = merge(var.tags, map("Name", format("Nessus%d", count.index)))
+}
--- a/nessus_ec2.tf
+++ b/nessus_ec2.tf
@@ -44,8 +44,7 @@ resource "aws_instance" "nessus" {
   user_data_base64 = data.template_cloudinit_config.nessus_cloud_init_tasks[count.index].rendered
 
   vpc_security_group_ids = [
-    aws_security_group.efs_client.id,
-    aws_security_group.operations.id,
+    aws_security_group.operations.id
   ]
 
   tags        = merge(var.tags, map("Name", format("Nessus%d", count.index)))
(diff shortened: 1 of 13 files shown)


nessus_ec2.tf
```diff
@@ -0,0 +1,53 @@
...
+  root_block_device {
+    volume_type           = "gp2"
+    volume_size           = 128
+    delete_on_termination = true
+  }
+
+  user_data_base64 = data.template_cloudinit_config.nessus_cloud_init_tasks[count.index].rendered
+
+  vpc_security_group_ids = [
+    aws_security_group.efs_client.id,
```
jsf9k: You put the Nessus instances in this security group, so I assume you intend for it to be able to mount the EFS file share.  If that is true, then you need to add a policy to allow that to the instance role.  See [here](https://github.com/cisagov/cool-assessment-terraform/blob/develop/kali_iam.tf#L36-L43) and [here](https://github.com/cisagov/cool-assessment-terraform/blob/develop/kali_iam.tf#L60-L81) for an example.  You will also need to add some cloud-init foo that is similar to [this](https://github.com/cisagov/cool-assessment-terraform/blob/develop/kali_cloud_init.tf#L13-L23) to make sure the mount point is created and the EFS file system is added to `/etc/fstab`.

If you _do not_ intend the Nessus instances to mount the EFS file share then you should remove them from this security group.

nessus_parameterstorereadonly_policy.tf
```diff
@@ -0,0 +1,25 @@
...
+# ------------------------------------------------------------------------------
+
+data "aws_iam_policy_document" "nessus_parameterstorereadonly_doc" {
+  statement {
+    actions = [
+      "ssm:GetParameter",
+      "ssm:GetParameters"
+    ]
+    resources = [
+      format("arn:aws:ssm:*:%s:parameter%s", local.images_account_id, var.ssm_key_nessus_admin_username),
```
jsf9k: `arn:aws:ssm:${local.images_account_id}:parameter${var.ssm_key_nessus_admin_username}` will also work here.  I'm fine with either syntax; I just wanted to point out that `format()` isn't required here.

jsf9k (changes requested):
You have either accidentally assigned the Nessus instances to the EFS client security group, or (more likely) not completely configured the Nessus instances to access the EFS file share.

felddy (approved):
Looks good.  Nice registration script.  See @jsf9k's comments about the EFS mount.

nessus_ec2.tf
dav3r: @jsf9k - I did not intend to give the Nessus instances access to the EFS share.  That was a copy/paste mistake on my part.

My reasoning was that if Nessus is only being accessed via the web GUI (from the Kali instances), there is no need for the Nessus instance to ever use the EFS share.  Can you think of any good reasons why it might be needed?

jsf9k: No, I just wasn't sure which way you were going with it.

dav3r: Thanks for catching this- I removed the Nessus instances from the EFS security group in https://github.com/cisagov/cool-assessment-terraform/pull/31/commits/b41b84f766d3b0ab892cfc252743434302dab332.

nessus_parameterstorereadonly_policy.tf
dav3r: I like the non-`format()` syntax a bit more, so I made the change in https://github.com/cisagov/cool-assessment-terraform/pull/31/commits/2abd89edf316433eff25bfb4d5ecbed3c43dc8c4.  Anything that is more "f-string-like" is good by me!  Thanks for mentioning this. 👍

jsf9k (approved):

PR #77, opened 2020-10-19
Bump up the instance type for Kali instances so they have more RAM

jsf9k:
## 🗣 Description

This pull request bumps up the instance type for Kali instances so they have more RAM.

## 💭 Motivation and Context

One of the fed leads asked for more RAM for the Kali instances.

## 🧪 Testing

This change is currently deployed to env7 in our production environment.

## 🚥 Types of Changes

- [x] Parameter tweak

## ✅ Checklist


jsf9k:
I don't know yet if this is a change we will want to adopt permanently, but I went ahead and added @dav3r as a reviewer just to keep him in the loop.

dav3r:
> I don't know yet if this is a change we will want to adopt permanently, but I went ahead and added @dav3r as a reviewer just to keep him in the loop.

Do we know what they are doing that requires more RAM?  It would be good to document that here.

jsf9k:
> > I don't know yet if this is a change we will want to adopt permanently, but I went ahead and added @dav3r as a reviewer just to keep him in the loop.
> 
> Do we know what they are doing that requires more RAM? It would be good to document that here.

They were guessing that Burp Suite was eating up all the RAM and that's why they saw a Kali instance lock up.  I'm not convinced they identified the root cause.  I think this was just a stab in the dark.

dav3r:
> > > I don't know yet if this is a change we will want to adopt permanently, but I went ahead and added @dav3r as a reviewer just to keep him in the loop.
> > 
> > 
> > Do we know what they are doing that requires more RAM? It would be good to document that here.
> 
> They were guessing that Burp Suite was eating up all the RAM and that's why they saw a Kali instance lock up. I'm not convinced they identified the root cause. I think this was just a stab in the dark.

OK, thanks for the context.  I don't think we should roll out this larger instance type across the board until we are sure it is truly needed.

jsf9k:
Fixes cisagov/cool-system#106.

dav3r (approved):
Give the people what they want! ⚡

hillaryj (approved):
🧑‍🤝‍🧑

mcdonnnj (approved):
image

PR #86, opened 2020-11-17
Make all instances that mount the EFS volume depend on the EFS mount target(s)

### LINKED ISSUE #85: EFS mount inconsistent with PentestPortal EC2 instances
# 🐛 Bug Report

When PentestPortal EC2 instances are spun up, the EFS mount sometimes fails.

## To Reproduce

- Spin up several PentestPortal EC2 instances
- Check to see if any have an attempted EFS mount that failed

## Expected behavior

The EFS mounts should always be available on PentestPortal instances.

jsf9k:
## 🗣 Description ##

This pull request adds a `depends_on = [aws_efs_mount.target.target]` key to all EC2 instances that mount the EFS volume.

## 💭 Motivation and Context ##

Issue #85 captures a few cases, particularly among the Pentest Portal instances, where the EFS volume was not mounted at boot.  Afterwards a simple `mount -a` command does successfully mount the EFS volume, the the issue was clearly not that the EC2 instances lacked the appropriate permissions.

I was eventually able to reproduce the problem.  I found that, when creating both the EFS mount target(s) and the EC2 instances, it was possible for the EFS mount target(s) to be created _after_ the EC2 instances that require it (them) to mount the EFS volume.  Adding a `depends_on = [aws_efs_mount.target.target]` key to all EC2 instances that mount the EFS volume ensures that the mount target(s) is (are) created _before_ such instances.

I am confident that this behavior was at the root of the bug reports we received, and therefore this pull request resolves #85.

## 🧪 Testing ##

As described above, I reproduced the error that was reported and then verified that the changes in this pull request corrected the error.

## ✅ Checklist ##

### DIFF
--- a/debiandesktop_ec2.tf
+++ b/debiandesktop_ec2.tf
@@ -25,8 +25,11 @@ data "aws_ami" "debiandesktop" {
 
 # The "Debian desktop" EC2 instances
 resource "aws_instance" "debiandesktop" {
-  count    = lookup(var.operations_instance_counts, "debiandesktop", 0)
-  provider = aws.provisionassessment
+  count = lookup(var.operations_instance_counts, "debiandesktop", 0)
+  # These instances require the EFS mount target to be present in
+  # order to mount the EFS volume at boot time.
+  depends_on = [aws_efs_mount_target.target]
+  provider   = aws.provisionassessment
 
   ami                         = data.aws_ami.debiandesktop.id
   associate_public_ip_address = true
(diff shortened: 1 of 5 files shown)


mcdonnnj (approved):
This makes sense to me. Nice work!

dav3r (approved):
Brilliant sleuthing Holmes! 🕵️ 
Nice informative comment too. 👍

PR #93, opened 2021-01-21
Add VPC endpoints

### LINKED ISSUE #94: Use VPC gateway and interface endpoints to remove HTTPS access and resources where possible
# 💡 Summary #

The Guacamole server in the private subnet of an assessment environment requires permission to send outgoing HTTPS requests in order to:
1. Obtain temporary credentials for its instance role via STS, and
2. Download its TLS certificate from an S3 bucket

It should be possible to do away with the external internet access and instead access STS and S3 via a [VPC interface endpoint](https://docs.aws.amazon.com/vpc/latest/userguide/integrated-services-vpce-list.html) and a [VPC gateway endpoint](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-endpoints-s3.html), respectively.

We should really be using these sorts of endpoints everywhere we can, since it keeps some of our AWS traffic from ever leaving AWS.

## Motivation and context ##

This is a good idea because:
* It will keep some of our AWS traffic from ever leaving AWS, and therefore increase security
* Once we do that we will no longer require certain NAT gateway resources
* By virtue of the previous bullet, it will simplify our OmniGraffle network diagrams for the COOL.  (@cisagov/team-ois - does anyone have a link to these diagrams in Google Docs?)

## Acceptance criteria ##

- [x] The STS interface endpoint is created
- [x] The S3 gateway endpoint is created
- [x] Outgoing access to the internet via HTTPS is removed, and the Guacamole server is still able to perform its functions.
- [x] All cloud-init scripts are able to complete successfully.
- [x] The staging environment is verified to function just as it did before.

jsf9k:
## 🗣 Description ##

This pull request:
* Adds several VPC endpoints for various AWS services
* Configures the operations and private subnets to make use of said VPC endpoints
* Shares a single VPC route table among the private subnets.  Since the private subnets do not each use a different NAT gateway in a different operations subnet, there is no need for them to have separate VPC route tables.

(It would be nice to remove the NAT gateway from the operations subnet, since the private subnets no longer require it to communicate with AWS services, but they still require it to pull the Docker images from Docker Hub for the Guacamole Docker composition.)

Resolves #94.

## 🧪 Testing ##

I deployed these changes to env0 in our staging COOL environment and verified that the cloud-init scripts, CloudWatch agent, and SSM agent are all running without errors on all private and operations instances.  I also verified that the FreeIPA initialization for the guacamole instance still works.

I also redeployed with the external route removed from the operations subnet's route table and verified that everything still worked, with the exception of the downloading of the Docker images required for the Guacamole Docker composition.  This serves as an ironclad check that all AWS API calls are being made to the VPC endpoints and none are being made to the public endpoints.

## ✅ Checklist ##

* [x] _All_ future TODOs are captured in issues, which are referenced in code comments.


dav3r:
> * Removes the NAT gateway from the operations subnet, since the private subnets no longer requires it to communicate with AWS services.

The description of this PR should be amended to remove this bullet, due to https://github.com/cisagov/cool-assessment-terraform/pull/93/commits/597a54164eb3dfbdd6e46e0bc1c8c70c0d50e46f. 

Also, doesn't this mean that https://github.com/cisagov/cool-assessment-terraform/pull/93/commits/a16b36a9756af5314b904c7a1e64bb4532b15d8f needs to be reverted as well?

jsf9k:
> > * Removes the NAT gateway from the operations subnet, since the private subnets no longer requires it to communicate with AWS services.
> 
> The description of this PR should be amended to remove this bullet, due to [597a541](https://github.com/cisagov/cool-assessment-terraform/commit/597a54164eb3dfbdd6e46e0bc1c8c70c0d50e46f).
> 
> Also, doesn't this mean that [a16b36a](https://github.com/cisagov/cool-assessment-terraform/commit/a16b36a9756af5314b904c7a1e64bb4532b15d8f) needs to be reverted as well?

I amended the PR description.

The commit a16b36a _does not_ need to be reverted.  As it turns out, the assessment VPCs were never set up like the shared services VPC with `N` public and `N` private subnets spread across AZs.  In that case we have each private subnet route its external traffic to the public subnet in the same AZ, so it makes sense to have separate route tables for each private subnet.  In this case the idea of separate route tables was implemented but not the more important idea of cross-AZ redundancy.

Sharp eyes!

private_routing.tf
```diff
@@ -12,46 +12,51 @@
...
 # -------------------------------------------------------------------------------
 
-# Each private subnet gets its own routing table, since each subnet
-# uses its own NAT gateway.
-resource "aws_route_table" "private_route_tables" {
+# The private subnets can all share a single routing table.  Normally
+# we would assign each subnet its own NAT gateway, which would require
+# separate routing tables, but in this case we do not create any NAT
+# gateways since nothing in the private subnets requires access to the
+# internet outside the VPC.
```
dav3r: This comment should be updated- the Guacamole instance in the private subnet still requires internet access to download the Guacamole Docker images and we still have the single NAT gateway used by the private subnets.

sts_endpoint_sg_rules.tf
```diff
@@ -0,0 +1,27 @@
...
+  source_security_group_id = aws_security_group.desktop_gateway.id
+  from_port                = 443
+  to_port                  = 443
+}
+
+# Allow ingress via HTTPS from the operations security group
+#
+# TODO - Only the Nessus instances require this access.  We could
+# avoid giving it to the other instance types by creating a separate
+# security group for Nessus instances.  See #95 for more details.
```
hillaryj: Do we want to include the link or just the number? It seems like we usually include a full link.

```suggestion
# security group for Nessus instances.  See #95 for more details:
# https://github.com/cisagov/cool-assessment-terraform/issues/95
```

vpc_endpoints.tf
```diff
@@ -0,0 +1,226 @@
...
+  # subnet_id       = aws_subnet.private[each.value].id
+  subnet_id       = aws_subnet.private[var.private_subnet_cidr_blocks[0]].id
+  vpc_endpoint_id = aws_vpc_endpoint.ec2.id
+}
+resource "aws_vpc_endpoint_subnet_association" "ec2messages" {
+  provider = aws.provisionassessment
+
+  # Normally we would use a for_each here, but the private subnets are
+  # _all in the same AZ_ in this case.  Why even have multiple subnets
+  # if you're not going to spread them around?  Harumph!
```
hillaryj, line 163: Harumph! https://www.youtube.com/watch?v=LmnaeU0k80I

hillaryj (approved):
One suggestion which is only a comment change, and otherwise: 🏆

dav3r (approved):
Excellent work and really great job on the comments!  🐎 

...except for one comment that needs a slight revision.

private_routing.tf
jsf9k: See commit 5a40a3a.

vpc_endpoints.tf
jsf9k: Also this bit from "Blazing Saddles":
[Harumph](https://www.youtube.com/watch?v=Yt3GBlVjUd0)

private_acl_rules.tf
```diff
@@ -139,7 +139,7 @@ resource "aws_network_acl_rule" "private_egress_to_operations_via_vnc" {
   network_acl_id = aws_network_acl.private[each.value].id
   egress         = true
   protocol       = "tcp"
-  rule_number    = 120 + index(var.private_subnet_cidr_blocks, each.value)
+  rule_number    = 130 + index(var.private_subnet_cidr_blocks, each.value)
```
felddy, line 157: 🧙 Seems magical.  Is this counting by 10 jazz useful?  Does this break if index ≥ 10?

subnets.tf
```diff
@@ -45,3 +45,22 @@ resource "aws_internet_gateway" "assessment" {
...
+# Create a NAT gateway for the private subnets.
+# -------------------------------------------------------------------------------
+resource "aws_eip" "nat_gw" {
+  provider = aws.provisionassessment
+
+  tags = var.tags
+  vpc  = true
+}
+
+resource "aws_nat_gateway" "nat_gw" {
```
felddy: Should we think about including the Guacamole Docker images as part of the AMI so that we can re-remove this gateway?

private_acl_rules.tf
jsf9k: Yes, it breaks if `index >= 10`.  Our numbering of the ACL rules has always sucked.  Do you have any thoughts as to how it can be improved?

cloud-init/install-certificates.py
```diff
@@ -24,7 +25,25 @@
...
+# set the region.  We have to set the region-specific endpoint URL
+# manually.
+#
+# This is important since the STS VPC endpoint _only_ sets a local DNS
+# record to override the _local region's_ public STS endpoint.  If we
+# don't set the endpoint URL then boto3 will reach out to the _global_
+# https://sts.amazonaws.com URL, and that DNS entry will still point
+# to an external IP.
+#
+# See this link for more information about boto3's perverse behavior
```
mcdonnnj, line 40: > boto3's perverse behavior

🤣

cloud-init/render-guac-connection-sql-template.py
```diff
@@ -31,7 +32,23 @@
 SSM_KEY_VNC_USER_PRIVATE_SSH_KEY = "${ssm_key_vnc_user_private_ssh_key}"
 
 # Create STS client
-sts = boto3.client("sts")
+#
```
felddy, line 35: Good catch!  👁️ 💪

subnets.tf
jsf9k: That is not a bad idea, but I have a sense of deja vu.  I think we discussed this before but decided against it, maybe because this was back when we were first cloudifying CyHy and making our AMIs public, and hence we wanted to keep sensitive data out of them.

@dav3r and @felddy, do you remember when we discussed this before?

felddy (approved):
Wonderful.  This makes a lot of sense. 💯 
I had a question about preloading the Docker images into the Guac AMI so that we can remove the NAT gateway.  We should discuss that. 🐳 🥑 

> My motto was always to keep swinging. Whether I was in a slump or feeling badly or having trouble off the field, the only thing to do was keep swinging. --Hank Aaron ⚾

subnets.tf
dav3r: I don't remember the details of this discussion from the past.  I am definitely in favor of baking the Guacamole Docker image into our AMI if it means we can get rid of the NAT gateway and related port 443 access rules.

jsf9k: OK.  I created cisagov/cool-system#158 for this change.

operations_acl_rules.tf
```diff
@@ -30,6 +30,23 @@ resource "aws_network_acl_rule" "operations_ingress_from_private_via_vnc" {
   to_port        = 5901
 }
 
+# Allow ingress from the private subnets via port 443.  This is
+# necessary so that the guacamole instance can download the Docker
```
mcdonnnj:
```suggestion
# necessary so that the Guacamole instance can download the Docker
```

operations_acl_rules.tf
```diff
@@ -30,6 +30,23 @@ resource "aws_network_acl_rule" "operations_ingress_from_private_via_vnc" {
...
+# images used in the docker composition via the NAT gateway.
```
mcdonnnj:
```suggestion
# images used in the Docker composition via the NAT gateway.
```

operations_acl_rules.tf
```diff
@@ -150,6 +167,10 @@ resource "aws_network_acl_rule" "operations_ingress_from_anywhere_via_icmp" {
 
 # Allow egress to anywhere via any protocol and port
 # For: Assessment team operational use
+#
+# Note that this also covers the return traffic when the guacamole
```
mcdonnnj:
```suggestion
# Note that this also covers the return traffic when the Guacamole
```

operations_acl_rules.tf
```diff
@@ -150,6 +167,10 @@ resource "aws_network_acl_rule" "operations_ingress_from_anywhere_via_icmp" {
...
+# instance downloads the Docker images used in the docker composition
```
mcdonnnj:
```suggestion
# instance downloads the Docker images used in the Docker composition
```

private_acl_rules.tf
```diff
@@ -15,7 +15,14 @@ resource "aws_network_acl_rule" "private_ingress_from_cool_via_https" {
 }
 
 # Allow egress to anywhere via HTTPS
-# For: Guacamole fetches its SSL certificate via boto3 (which uses HTTPS)
+#
+# For: Guacamole fetches its SSL certificate via boto3 (which uses
+# HTTPS).  It also needs to download the Docker images used in the
+# guacamole Docker composition.
```
mcdonnnj:
```suggestion
# Guacamole Docker composition.
```

private_acl_rules.tf
```diff
@@ -15,7 +15,14 @@ resource "aws_network_acl_rule" "private_ingress_from_cool_via_https" {
 }
 
 # Allow egress to anywhere via HTTPS
-# For: Guacamole fetches its SSL certificate via boto3 (which uses HTTPS)
+#
+# For: Guacamole fetches its SSL certificate via boto3 (which uses
```
mcdonnnj: This is now inconsistent with the changes made here:
https://github.com/cisagov/cool-assessment-terraform/blob/5a40a3a80152600df6cad2d489a47f27eae17711/desktop_gateway_sg_rules.tf#L30-L31
and here:
https://github.com/cisagov/cool-assessment-terraform/blob/5a40a3a80152600df6cad2d489a47f27eae17711/desktop_gateway_sg_rules.tf#L75-L76

vpc_endpoints.tf
```diff
@@ -0,0 +1,226 @@
...
+  # for_each = toset(var.private_subnet_cidr_blocks)
+
+  # subnet_id       = aws_subnet.private[each.value].id
+  subnet_id       = aws_subnet.private[var.private_subnet_cidr_blocks[0]].id
+  vpc_endpoint_id = aws_vpc_endpoint.monitoring.id
+}
+
+# S3 gateway endpoint
+resource "aws_vpc_endpoint" "s3" {
```
mcdonnnj, line 227: Is there a reason this isn't with the other `aws_vpc_endpoint` declarations?

desktop_gateway_sg_rules.tf
```diff
@@ -24,6 +25,66 @@ resource "aws_security_group_rule" "desktop_gw_egress_to_anywhere_via_https" {
...
+  security_group_id = aws_security_group.desktop_gateway.id
+  type              = "egress"
+  protocol          = "tcp"
+  prefix_list_ids   = [aws_vpc_endpoint.s3.prefix_list_id]
+  from_port         = 443
+  to_port           = 443
+}
+
 # Allow ingress from COOL Shared Services VPN server CIDR block
 # via port 443 (nginx/guacamole web)
```
mcdonnnj:
```suggestion
# via port 443 (nginx/Guacamole web)
```

mcdonnnj (approved):
This is a pretty big lift and some strong work 💪💪💪
I had some suggestions/questions, but I was also curious if we have a standard on capitalization of protocols? We seem to use for example `https` and `HTTPS` without clear purpose.

private_acl_rules.tf
jsf9k: I'm not following how it is inconsistent.  Can you elaborate?

Remember that with VPC _gateway_ endpoints (used only with S3 and DynamoDB), the traffic leaves the subnet destined for _the real, external IP_ to which `s3.[region.]amazonaws.com` resolves but is directed to the VPC gateway endpoint by the route table.  Hence the subnet's ACL needs to allow HTTPS egress anywhere.

jsf9k:
> I was also curious if we have a standard on capitalization of protocols? We seem to use for example `https` and `HTTPS` without clear purpose.

We do not have such a standard, but you are welcome to propose one.  Capitalization is probably the way to go.

vpc_endpoints.tf
jsf9k: See commit 2d42f57.

private_acl_rules.tf
mcdonnnj: Sorry, I just meant the comment language changed, and this comment does not match the change.
`Guacamole fetches its SSL certificate via boto3 (which uses HTTPS)` -> `Guacamole requires access to S3 in order to download its certificate.`

vpc_endpoints.tf
mcdonnnj: I thought that might have been why, but the comments make it explicit. Thanks!

mcdonnnj:
> 
> 
> > I was also curious if we have a standard on capitalization of protocols? We seem to use for example `https` and `HTTPS` without clear purpose.
> 
> We do not have such a standard, but you are welcome to propose one. Capitalization is probably the way to go.

If I'm proposing, then yes, I would vote for defaulting to capitalization for protocols.

jsf9k:
> > > I was also curious if we have a standard on capitalization of protocols? We seem to use for example `https` and `HTTPS` without clear purpose.
> > 
> > 
> > We do not have such a standard, but you are welcome to propose one. Capitalization is probably the way to go.
> 
> If I'm proposing, then yes, I would vote for defaulting to capitalization for protocols.

You could add something to [this long-running PR](https://github.com/cisagov/development-guide/pull/44).

private_acl_rules.tf
jsf9k: See commit d9efdca.

PR #112, opened 2021-04-15
Add and use email-sending domain

### LINKED ISSUE #110: Add and use an email-sending domain input variable
# 💡 Summary #

Make an input variable for the domain where emails will be sent from and configure the Postfix container on the Gophish instance to correctly use that domain.

## Motivation and context ##

Many of our assessments have the need to send emails from a specific domain, which currently requires several manual configuration steps.  This would automate part of that process.

This is part of https://github.com/cisagov/cool-system/issues/90.

## Implementation notes ##

Ensure that the `PRIMARY_DOMAIN` is set for Postfix before the container starts up.  As long as this is done by cloud-init, I think we should be fine, since the pca-gophish-composition SystemD service [doesn't start up until after `cloud-final.service`](https://github.com/cisagov/pca-gophish-composition-packer/blob/develop/src/pca-gophish-composition.service#L3).

## Acceptance criteria ##

When a fresh assessment environment containing a Gophish instance is deployed:
- [ ] The Postfix container's primary domain is correctly set based on the value of the Terraform input variable for the email-sending domain
- [ ] The Postfix container and the pca-gophish-composition SystemD service on the Gophish instance start up correctly
- [ ] Email can be successfully sent via Postfix from the email-sending domain

dav3r:
#  #

## 🗣 Description ##
This PR adds:
* A new Terraform input variable for the email-sending domain to be used in assessments
* A cloud-init script that sets the `PRIMARY_DOMAIN` variable for the postfix service in the PCA-Gophish Docker composition to the value of the Terraform email-sending domain variable

As specified in #110:
> Many of our assessments have the need to send emails from a specific domain, which currently requires several manual configuration steps. This would automate part of that process.

Resolves #110.

## 🧪 Testing ##

I applied these Terraform changes in a COOL environment and confirmed that:
* The Postfix container's primary domain was correctly set based on the value of the Terraform input variable for the email-sending domain.
* The Postfix container and the pca-gophish-composition SystemD service on the Gophish instance started up correctly.
* Email was successfully sent via Postfix in the assessment environment from the email-sending domain.

## ✅ Checklist ##

* [x] All relevant repo and/or project documentation has been updated


README.md
```diff
@@ -95,6 +95,7 @@ the COOL environment.
 | cert_bucket_name | The name of the AWS S3 bucket where certificates are stored. | `string` | `cisa-cool-certificates` | no |
 | cool_domain | The domain where the COOL resources reside (e.g. "cool.cyber.dhs.gov"). | `string` | `cool.cyber.dhs.gov` | no |
 | dns_ttl | The TTL value to use for Route53 DNS records (e.g. 86400).  A smaller value may be useful when the DNS records are changing often, for example when testing. | `number` | `60` | no |
+| email_sending_domain | The domain to send emails from within the assessment environment (e.g. "example.com"). | `string` | `example.com` | no |
```
jsf9k, line 98: Do you mean the domain or the hostname?  If the user wants to send email from `treeofwoe.thulsa-doom.com` is that the value they specify for this variable?  Or do they just specify the domain `thulsa-doom.com` and the `treeofwoe` is specified elsewhere?

hillaryj (approved):
🗡️

cloud-init/postfix-setup.tpl.yml
```diff
@@ -0,0 +1,6 @@
+---
+runcmd:
+  # Modify postfix's PRIMARY_DOMAIN in pca-gophish-composition docker-compose
+  # file to the email-sending domain for this assessment
+  - [sed, -i, "s|PRIMARY_DOMAIN=.*|PRIMARY_DOMAIN=${email_sending_domain}|g",
```
jsf9k: This regex will apply to lines that look like `.*PRIMARY_DOMAIN=.*$`.  I suspect that you want it to apply only to lines that look like `^PRIMARY_DOMAIN=[^\w]*$`.  I suggest changing the regex to this to be extra safe:
```suggestion
  - [sed, -i, "s|^PRIMARY_DOMAIN=[^\w]*$|PRIMARY_DOMAIN=${email_sending_domain}|g",
```

jsf9k (changes requested):
I have a couple of questions.

README.md
jsf9k: If the user is supposed to specify the hostname then I suggest changing the name of the variable to make this clearer.  At a minimum, you should change the default to `mail.example.com` if the user is expected to specify the hostname instead of just the domain.

cloud-init/postfix-setup.tpl.yml
mcdonnnj: Based on the `docker-compose.yml` this is likely editing, at least part of your edit would not work: https://github.com/cisagov/pca-gophish-composition/blob/47c69034a5df627852444932ed960660aeadce2b/docker-compose.yml#L67

jsf9k: Ah, your eyes are as sharp as ever @mcdonnnj!  How about this then:
```suggestion
  - [sed, -i, "s|^\([- \w]*\)PRIMARY_DOMAIN=[^\w]*$|\1PRIMARY_DOMAIN=${email_sending_domain}|g",
```

README.md
dav3r: Honestly, I don't know how much this value really matters.  If you set it to `thulsa-doom.com`, you can successfully send mail from either `conan@thulsa-doom.com` or `conan@treeofwoe.thulsa-doom.com`.  I did a test of this and it looks like as long as there is a valid MX record for `thulsa-doom.com` and `treeofwoe.thulsa-doom.com`, either one should be acceptable.

That is why I worded it as `domain` instead of `hostname` or `fully-qualified domain name`.

PS: I used https://mail-tester.com to get a score (with explanations) for the test emails I sent and they were scored the same, regardless of whether they were sent from `conan@thulsa-doom.com` or `conan@treeofwoe.thulsa-doom.com`.  For these tests, my `email_sending_domain` was just the `thulsa-doom.com` domain (not really, but the domain that I have been using for my testing).

jsf9k: Perfect.  I withdraw the question.

cloud-init/postfix-setup.tpl.yml
dav3r: Thanks @mcdonnnj, you beat me to the punch on the `docker-compose.yml` stuff. 😄  

@jsf9k, your most-recent suggestion is almost perfect, but I had to escape the backslashes to appease the Python YAML linter used by cloud-init as well as add a `\` linebreak to satisfy our pre-commit `yamllint` hook.  

I verified that it worked as expected by deploying it to my test environment, then made the change in https://github.com/cisagov/cool-assessment-terraform/pull/112/commits/4d2c4c29757ea4fabfd570a78323e2f40c78408b.

Thanks for the improvement!

jsf9k (approved):
Great work!  Thanks for performing a real test of the suggested regex changes.  I tested locally by `echo`ing various strings and piping that into `sed`, but a real test is vastly better.

dav3r:
> Great work! Thanks for performing a real test of the suggested regex changes. I tested locally by `echo`ing various strings and piping that into `sed`, but a real test is vastly better.

FYI, after a couple of rounds of trying to fix the YAML errors (requiring a redeploy of my Gophish instance each time), I discovered [this handy bit of Python from cloud-init](https://github.com/canonical/cloud-init/blob/master/tools/validate-yaml.py), which allowed me to quickly verify locally that all was well.  After I got that `validate-yaml.py` script to confirm that my YAML was valid, I redeployed my instance with it and confirmed that everything was groovy.

PR #142, opened 2021-10-13
Make Terraformer access to IAM read-only

jsf9k:
## 🗣 Description ##

This pull requests adds a statement to the Terraformer policy document that makes its IAM access read-only.

This stops the Terraformer instance from defeating the Terraformer policy by creating new users, policies, roles, etc.  I'm looking at you, @epicfaace!  :smiley:

## 🧪 Testing ##

I successfully applied these changes to env1 in our staging COOL environment.

## ✅ Checklist ##

### DIFF
--- a/terraformer_policy.tf
+++ b/terraformer_policy.tf
@@ -14,8 +14,8 @@ data "aws_iam_policy_document" "terraformer_policy_doc" {
   # root module, with the exception of IAM.
   #
   # We will attach the arn:aws:iam::aws:policy/ReadOnlyAccess policy
-  # to the same role to which policy document will be attached, which
-  # will give is read-only access to all IAM resources.  We require
+  # to the same role to which the policy document will be attached, which
+  # will give it read-only access to all IAM resources.  We require
   # IAM access to be as read-only as possible in order to stop the
   # Terraformer instance from defeating the Terraformer policy by
   # creating new users, policies, roles, etc.
--- a/terraformer_policy.tf
+++ b/terraformer_policy.tf
@@ -11,11 +11,14 @@ data "aws_iam_policy_document" "terraformer_policy_doc" {
 
   # Allow full access to new resources and existing resources that
   # _are not_ tagged as being created by the team that deploys this
-  # root module.
+  # root module, with the exception of IAM.
+  #
+  # We will attach the arn:aws:iam::aws:policy/ReadOnlyAccess policy,
+  # which will cover the IAM case.  We require IAM access to be
+  # read-only (as far as is possible) in order to stop the Terraformer
+  # instance from defeating the Terraformer policy by creating new
+  # users, policies, roles, etc.
   statement {
-    actions = [
-      "*",
-    ]
     condition {
       test = "StringNotEquals"
       values = [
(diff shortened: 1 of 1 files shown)


dav3r (approved):
Makes sense to me! 🔒

terraformer_policy.tf
```diff
@@ -28,6 +28,23 @@ data "aws_iam_policy_document" "terraformer_policy_doc" {
     ]
   }
 
+  # Deny all IAM access unless it is read-only.  This stops the
+  # Terraformer instance from defeating the Terraformer policy by
+  # creating new users, policies, roles, etc.
+  #
+  # I cribbed this list of actions from the AWS built-in policy
+  # arn:aws:iam::aws:policy/IAMReadOnlyAccess.
```
mcdonnnj: I think we should mention the rationale for not including the `iam:Generate*` permission from the built-in policy. An alternative would be to link to [this page in the AWS documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_examples_iam_read-only-console-no-reporting.html) since it matches the permissions we're allowing.

mcdonnnj (changes requested):
Makes sense. I had one request related to clarifying the documentation around the new policy statement.

terraformer_policy.tf
jsf9k: I believe this outdated conversation can now be resolved.

terraformer_policy.tf
```diff
@@ -11,22 +11,58 @@ data "aws_iam_policy_document" "terraformer_policy_doc" {
 
   # Allow full access to new resources and existing resources that
   # _are not_ tagged as being created by the team that deploys this
-  # root module.
+  # root module, with the exception of IAM.
+  #
+  # We will attach the arn:aws:iam::aws:policy/ReadOnlyAccess policy,
+  # which will cover the IAM case.  We require IAM access to be
```
dav3r: This comment now needs to be updated.

dav3r (approved):
Other than the comment change that I noted, this all looks good to me. 🦘

terraformer_policy.tf
jsf9k: That policy gets attached [here](https://github.com/cisagov/cool-assessment-terraform/blob/93c422af1983953f184a70a7f2f5f5000747da87/terraformer_role.tf#L13-L19).  What needs to change?

jsf9k: Have a look at commit 041d892 and let me know if that scratches your itch.

dav3r: > That policy gets attached [here](https://github.com/cisagov/cool-assessment-terraform/blob/93c422af1983953f184a70a7f2f5f5000747da87/terraformer_role.tf#L13-L19). What needs to change?

Arg, I totally overlooked that- this is why I shouldn't review PRs at the end of the day.  Though the enhanced comments in https://github.com/cisagov/cool-assessment-terraform/commit/041d892a59aaa6bafa6a6b1fa0ccede95837f0b8 are lovely. 👍

terraformer_policy.tf
```diff
@@ -11,22 +11,60 @@ data "aws_iam_policy_document" "terraformer_policy_doc" {
 
   # Allow full access to new resources and existing resources that
   # _are not_ tagged as being created by the team that deploys this
-  # root module.
+  # root module, with the exception of IAM.
+  #
+  # We will attach the arn:aws:iam::aws:policy/ReadOnlyAccess policy
+  # to the same role to which policy document will be attached, which
+  # will give is read-only access to all IAM resources.  We require
```
dav3r: Fix a couple of typos:
```suggestion
  # We will attach the arn:aws:iam::aws:policy/ReadOnlyAccess policy
  # to the same role to which the policy document will be attached, which
  # will give it read-only access to all IAM resources.  We require
```

terraformer_policy.tf
jsf9k: No problem sir!

mcdonnnj (approved):
This makes sense. 🔒

PR #143, opened 2021-10-15
Add rules to allow anything in the Operations subnet to use the VPC endpoints

jsf9k:
## 🗣 Description ##

This pull request adds rules to allow anything in the Operations subnet to use the VPC endpoints.

This is necessary since instances created via Terraformer live in the Operations subnet, but they do not have a security group that I can list in a SG rule.

## 🧪 Testing ##

I have applied these changes in env3 of our staging COOL environment and verified that they function as expected.

## ✅ Checklist ##

### DIFF
--- a/cloudwatch_endpoint_sg.tf
+++ b/cloudwatch_endpoint_sg.tf
@@ -117,3 +117,15 @@ resource "aws_security_group_rule" "ingress_from_terraformer_to_cloudwatch_via_h
   from_port                = 443
   to_port                  = 443
 }
+
+# Allow ingress via HTTPS from the operations subnet
+resource "aws_security_group_rule" "ingress_from_operations_subnet_to_cloudwatch_via_https" {
+  provider = aws.provisionassessment
+
+  security_group_id = aws_security_group.cloudwatch.id
+  type              = "ingress"
+  protocol          = "tcp"
+  cidr_blocks       = [var.operations_subnet_cidr_block]
+  from_port         = 443
+  to_port           = 443
+}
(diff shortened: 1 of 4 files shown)


cloudwatch_endpoint_sg.tf
```diff
@@ -117,3 +117,15 @@ resource "aws_security_group_rule" "ingress_from_terraformer_to_cloudwatch_via_h
...
+resource "aws_security_group_rule" "ingress_from_operations_subnet_to_cloudwatch_via_https" {
+  provider = aws.provisionassessment
+
+  security_group_id = aws_security_group.cloudwatch.id
+  type              = "ingress"
+  protocol          = "tcp"
+  cidr_blocks       = [var.operations_subnet_cidr_block]
+  from_port         = 443
+  to_port           = 443
+}
```
dav3r, line 131: Doesn't adding this rule mean that we can get rid of all of the other rules that are specific to instances that live in the Operations subnet?

https://github.com/cisagov/cool-assessment-terraform/blob/93c422af1983953f184a70a7f2f5f5000747da87/cloudwatch_endpoint_sg.tf#L13-L119

The same question applies to the new SSM and STS endpoint rules you added here.

jsf9k: Good question.  Now that I know that the number of security groups applied to an eni is not a hard limit, my intent is to come back and create a single `cloudwatch_endpoint_client_sg` and use that here.  That will allow me to drop this new CIDR-based rule and get rid of all the other SG-based rules.  I will do something similar for the other VPC endpoint security groups.  Instances spun up via the Terraformer instance can then simply add themselves to those security groups as necessary and gain access to the necessary VPC endpoints.

This change will also allow me to get rid of the existing instance type SGs, except for those that actually require an explicit mention.  For example, the Debian desktop SG will probably be rendered unnecessary as it isn't mentioned explicitly outside of these endpoint SGs, but the Teamserver instance will still be becessary because certain other need to be instance types are allowed ingress to the Teamservers.

dav3r: Understood, thanks for the additional information here.

dav3r (approved):
Fully Operational! 📶

jsf9k:
> Fully Operational! signal_strength

:new_moon: <-- That's no moon!

mcdonnnj (approved):
Makes sense ✔

cloudwatch_endpoint_sg.tf
mcdonnnj: @jsf9k Is there an issue to track that work? If not would you mind creating one?

jsf9k: See #144.

PR #154, opened 2021-12-02
Change ownership of the EFS share mount point

mcdonnnj:
#  #

## 🗣 Description ##

This PR adjusts the cloud-config logic for setting up the EFS share to `chown` the mount point using the provided owner and group.

Currently the mount point is created as the `root` user. This means you must have `root` privileges to access the share. On Kali machines this is fine because `sudo` permissions are configured, but since Windows instances must go through a Samba instance this creates a problem. Currently a Windows instance cannot write to the EFS share until a sub-directory with the appropriate `uid` (the `vnc` and `smbguest` `uid`) has been created. With this in place the Windows instance will be able to immediately write to the EFS share and other system users will not need to `sudo` to access the share.

## 🧪 Testing ##

Automated tests pass. I deployed this in my testing environment and confirmed that the share is being `chown`ed correctly and the Windows instance is able to write to the root EFS directory with no additional changes.

## ✅ Checklist ##

### DIFF
--- a/assessorportal_cloud_init.tf
+++ b/assessorportal_cloud_init.tf
@@ -18,7 +18,9 @@ data "cloudinit_config" "assessorportal_cloud_init_tasks" {
       "${path.module}/cloud-init/efs-mount.tpl.yml", {
         # Just mount the EFS mount target in the first private subnet
         efs_id      = aws_efs_mount_target.target[var.private_subnet_cidr_blocks[0]].file_system_id
+        group       = "vnc"
         mount_point = "/share"
+        owner       = "vnc"
     })
     content_type = "text/cloud-config"
     filename     = "efs_mount.yml"
--- a/assessorportal_cloud_init.tf
+++ b/assessorportal_cloud_init.tf
@@ -18,9 +18,9 @@ data "cloudinit_config" "assessorportal_cloud_init_tasks" {
       "${path.module}/cloud-init/efs-mount.tpl.yml", {
         # Just mount the EFS mount target in the first private subnet
         efs_id      = aws_efs_mount_target.target[var.private_subnet_cidr_blocks[0]].file_system_id
-        group       = "vnc"
+        group       = data.aws_ssm_parameter.vnc_username.value
         mount_point = "/share"
-        owner       = "vnc"
+        owner       = data.aws_ssm_parameter.vnc_username.value
     })
     content_type = "text/cloud-config"
     filename     = "efs_mount.yml"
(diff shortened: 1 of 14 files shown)


jsf9k (approved):

assessorportal_cloud_init.tf
```diff
@@ -18,7 +18,9 @@ data "cloudinit_config" "assessorportal_cloud_init_tasks" {
       "${path.module}/cloud-init/efs-mount.tpl.yml", {
         # Just mount the EFS mount target in the first private subnet
         efs_id      = aws_efs_mount_target.target[var.private_subnet_cidr_blocks[0]].file_system_id
+        group       = "vnc"
         mount_point = "/share"
+        owner       = "vnc"
```
dav3r: Not to complicate things, but we have tried to avoid hard-coding the username as `vnc`.  Instead, we should obtain the username by looking up the value of the SSM parameter specified by `var.ssm_key_vnc_username`.  I'm fine with you making a TODO issue to fix this later if you don't have time for it now.

For an example of how we do this, take a look at [`cloud-init/write-aws-config.tpl.sh`](https://github.com/cisagov/cool-assessment-terraform/blob/develop/cloud-init/write-aws-config.tpl.sh).

As long as we are doing this, it might also be a good idea to add the VNC group to SSM as well, though I'm fine assuming the username and group are the same for now.

dav3r (changes requested):
I've got one thang I'd like to see addressed before I approve.

assessorportal_cloud_init.tf
mcdonnnj: The only issue I see is that this is a cloud-config file and not a shell script. Thoughts on using an [`aws_ssm_parameter` data source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/ssm_parameter)?

dav3r: I don't love this bit with the `aws_ssm_parameter` data source:
```
The unencrypted value of a SecureString will be stored in the raw [Terraform] state as plain-text.
```

In this case, it wouldn't be that hard to turn this cloud-config file into a shell script.  Or maybe just split out the `runcmd` bits into their own shell script if we don't want to lose the simplicity of using that `mounts` command.

dav3r: PS: I'm not worried about this particular SSM value being in the Terraform state- just more generally about us going down that path and ending up with something more sensitive landing in the state.

mcdonnnj: Yeah I only suggested it given the non-sensitive nature of this value. We could always add a disclaimer that the value is non-sensitive so it's fine if it's in the Terraform state?

dav3r: I suppose that's ok in this case.  Though if it's not sensitive, then maybe it doesn't even belong in SSM?  Though it is convenient to have certain values in SSM if they are accessed by Terraform, cloud-init, and packer scripts.

jsf9k: > Though if it's not sensitive, then maybe it doesn't even belong in SSM?

SSM Parameter Store is for storing parameters, secret or otherwise.  There's nothing wrong with storing non-secret parameters there for convenience, although in many cases it may be simpler to store such parameters in, e.g. a `*.tfvars` file.

mcdonnnj: Please enjoy d0f2f37 for a non-static version.

dav3r (approved):
Looks fantastic now! 🥇

assessorportal_cloud_init.tf
dav3r: Perfecto - nice work.

jsf9k (approved):

PR #155, opened 2021-12-02
Use a consistent group for ownership of EFS mount points

mcdonnnj:
#  #

## 🗣 Description ##

This PR builds on the work in #154 by changing to use the group created by [cisagov/ansible-role-amazon-efs-utils] when changing ownership of the EFS mount point.

The default `efs_user` is a user group created by [cisagov/ansible-role-amazon-efs-utils] and should be present on every instance that would mount the EFS share. It makes more sense to use this existing user group when changing the ownership of EFS share mount point because it will also provide a consistent gid at the root of the EFS share. I am using a variable with a default because that matches how [cisagov/ansible-role-amazon-efs-utils] is configured to set the group name.

## 🧪 Testing ##

Automated tests pass. I deployed this change in my testing environment and observed that mount point group was correct on instance.

## ✅ Checklist ##

[cisagov/ansible-role-amazon-efs-utils]: https://github.com/cisagov/ansible-role-amazon-efs-utils


dav3r (approved):
Makes sense to me- onward! 🏇

jsf9k (approved):

PR #159, opened 2021-12-09
Use an EFS access point to enforce file permissions on the EFS share

### LINKED ISSUE #156: Control EFS share permissions using EFS access points
# 💡 Summary #

We should add [EFS access points](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html) to control file-system access to the EFS share.

## Motivation and context ##

Currently there is no mechanism to enforce file-system permissions on access to the EFS share. This means that files and directories with uids/gids different from those owning the EFS share mount point can be created. This can result in difficulties interacting with EFS share contents on systems that cannot `sudo` (like Windows instances).

## Implementation notes ##

There should be an access point for each EFS mount target created (there is only one currently). The `efs-mount` cloud-config will need to be updated to support this functionality as well.

## Acceptance criteria ##

How do we know when this work is done?

- [ ] EFS share file-system permissions are enforced

mcdonnnj:
#  #

## 🗣 Description ##

This PR adds an [EFS access point](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html) for each [EFS mount target](https://docs.aws.amazon.com/efs/latest/ug/accessing-fs.html) (currently only one). The access point is configured to enforce the UID and GID for file-system access to the EFS share.

Right now any resources in the EFS share will use the UID/GID of whatever user is creating them. This means that if a user on a Kali system adds resources to the EFS share as the `root` user they will have the the UID/GID of the `root` user. This can cause problems for interacting with resource on the EFS share for other systems (most notably Windows instances). Using the access point will ensure that any resources created in the EFS share have the UID of the `vnc`/`smbguest` users and the GID of the `efs_users` group. This will ensure that resource access is consistent and available for all users of the EFS share.

## 🧪 Testing ##

Automated tests pass. I applied this in my testing environment and everything worked as expected.

## 📷 Screenshots ##

image
image

## ✅ Checklist ##

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,7 @@ the COOL environment.
 | [aws_ec2_transit_gateway_route.assessment_route](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ec2_transit_gateway_route) | resource |
 | [aws_ec2_transit_gateway_route_table_association.association](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ec2_transit_gateway_route_table_association) | resource |
 | [aws_ec2_transit_gateway_vpc_attachment.assessment](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ec2_transit_gateway_vpc_attachment) | resource |
+| [aws_efs_access_point.access_point](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/efs_access_point) | resource |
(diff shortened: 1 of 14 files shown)


jsf9k (approved):
Strong work!  A simple solution for a long-running source of woe.

dav3r (approved):
Great work solving this so cleanly! 🥇

PR #160, opened 2022-01-05
Update the EFS access point configuration to resolve permissions issues

mcdonnnj:
#  #

## 🗣 Description ##

This pull request updates the [AWS EFS access point](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html) configuration to use a specified path in the EFS volume. This includes creating that path if it does not exist and setting expected permissions to mirror the permissions enforced by the access point.

A fresh EFS volume is owned by `root:root` and this can only be changed (that I've been able to find) by mounting the EFS without the access point and `chown`ing the mount point. This creates permissions errors because the access point coerces access to the mount point as a specific, non-root UID:GID. We can resolve this by creating a sub-directory in the EFS share that is mounted for the access point and configure that sub-directory to be created with permissions that match the access point configuration.

## 🧪 Testing ##

Automated tests pass. I confirmed that on a fresh deployment the `/share` mount has the correct permissions. I manually mounted the root EFS volume and confirmed that a directory was created for the access point.


efs.tf
```diff
@@ -43,6 +43,18 @@ resource "aws_efs_access_point" "access_point" {
     gid = var.efs_access_point_gid
     uid = var.efs_access_point_uid
   }
+
+  # Since we only use an access point to access the EFS share we need to make sure
+  # the automatically created root directory (on the EFS share) has the correct
+  # permissions or they will override the local configuration.
```
jsf9k:
```suggestion
  # permissions.
```

jsf9k (approved):
I made one small suggestion that you can take or leave, but this looks good to me.

dav3r (approved):
Excellent job figuring this out and fixing it! 🥇 

Someone could still be a dummy and set `efs_access_point_root_directory` to `"/"`, but unless we can think of a simple way to avoid that, I don't think we need to worry too much about this.

efs.tf
mcdonnnj: Do you have any thoughts on tweaking that vs outright removal? I'd like to keep mention that the local mount point's permissions will be overridden, which is a problem if they don't align.

jsf9k: I'm fine if it stays, but I find the "override the local configuration" language unclear.  I had to think about it to figure out what you were saying.  "...override the permissions on the local mount point" is a better substitute, so I'd be OK with using that language instead.

In any case, you can disregard if you disagree.  It isn't worth holding up the PR for this tiny point.

mcdonnnj: @jsf9k Would you look at abb95cd and let me know what you think?

jsf9k: Looks good!

PR #164, opened 2022-01-27
Require all email sending domains be lowercase

### LINKED ISSUE #162: Convert `email_sending_domains` to lowercase before use
# 🐛 Summary #

If any items in the `email_sending_domains` Terraform variable contain uppercase letters, their corresponding certificates will not be found in the S3 bucket because they are always lowercase in the bucket.

## To reproduce ##

1. Create a certificate for `xyz.com` and store it in the certificates bucket
1. Define `email_sending_domains = ["XYZ.com"] in a `.tfvars` file
1. Run a `terraform apply` with that `.tfvars` file

Note that the certificate cannot be pulled via our [`install-certificates.py` cloud-init](https://github.com/cisagov/cool-assessment-terraform/blob/develop/cloud-init/install-certificates.py) - it will fail with an error like this:

```
Error fetching 'mycertificatesbucket/live/XYZ.com/fullchain.pem' from S3: An error occurred (AccessDenied) when calling the GetObject operation: Access Denied
Exiting script!
```

## Expected behavior ##

The uppercase domain `XYZ.com` should be converted to lowercase before we attempt to fetch the certificate from the S3 bucket, so that `install-certificates.py` runs successfully.

CC: @m1j09830

dav3r:
#  #

## 🗣 Description ##

This PR ensures that any domains present in `var.email_sending_domains` do not have any uppercase letters in them.  This is accomplished by adding a [custom validation rule](https://www.terraform.io/language/values/variables#custom-validation-rules), as [brilliantly suggested by @jsf9k](https://github.com/cisagov/cool-assessment-terraform/pull/164#pullrequestreview-865377857).

From #162:
> If any items in the email_sending_domains Terraform variable contain uppercase letters, their corresponding certificates will not be found in the S3 bucket because they are always lowercase in the bucket.

This PR corrects that situation and thus resolves #162.

## 🧪 Testing ##

In an environment that was already deployed, I temporarily changed the `email_sending_domains` value in its Terraform variables file from all lowercase to all uppercase.  Then I ran a `terraform plan` and I confirmed that Terraform correctly displayed the following validation error:

```
│ Error: Invalid value for variable
│ 
│   on variables.tf line 94:
│   94: variable "email_sending_domains" {
│ 
│ All of the values in email_sending_domains must be lowercase.
```

I confirmed this worked with a few other mixes of uppercase and lowercase letters and with multiple domains where some included all lowercase and others included some uppercase.  I also verified that when all of the domains had no uppercase letters, the error message did not appear in that situation.


jsf9k (commented):
Should we make use of [custom validation rules](https://www.terraform.io/language/values/variables#custom-validation-rules) on input variables instead of using a local variable?  That seems like a simpler solution here.

dav3r:
> Should we make use of [custom validation rules](https://www.terraform.io/language/values/variables#custom-validation-rules) on input variables instead of using a local variable? That seems like a simpler solution here.

Check out the big brain on @jsf9k!  That's a great idea, thanks.  I'll try this out- stay tuned.

jsf9k (approved):
Great!

Before merging, we should also rebase and force-push to get rid of the first two commits and their reversions.  I can do that if you want.

dav3r:
> Great!
> 
> Before merging, we should also rebase and force-push to get rid of the first two commits and their reversions. I can do that if you want.

It's your call- I don't mind having them there to show how we got here, but if they bother you, go for it.

jsf9k:
> > Great!
> > Before merging, we should also rebase and force-push to get rid of the first two commits and their reversions. I can do that if you want.
> 
> It's your call- I don't mind having them there to show how we got here, but if they bother you, go for it.

I'll remove them, then.  They confuse the history, and there is no point in leaving the scaffolding up once the building is complete.

mcdonnnj (approved):
Makes sense and using validation is an excellent improvement (and something we should look at leveraging in more of our projects).

PR #177, opened 2022-03-16
Add standard CloudWatch instance alarms for all instance types

jsf9k:
## 🗣 Description ##

This pull request adds a set of standard CloudWatch instance alarms for all instance types.

These alarms will notify us via email if any of the following should occur:
- Either of an instance's status checks (instance or system) fail
- Any IMDSv1 requests succeed
- CPU, memory, or disk utilization is high
- Any packets are queued and/or dropped because the a networking limitation was exceeded for the instance

## 🧪 Testing ##

All automated testing passes.  This code is also already applied to env6 in our COOL staging environment.

## ✅ Pre-merge checklist ##

- [x] Revert dependencies to default branches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,7 @@ the COOL environment.
 | cw\_alarms\_samba | github.com/cisagov/instance-cw-alarms-tf-module | n/a |
 | cw\_alarms\_teamserver | github.com/cisagov/instance-cw-alarms-tf-module | n/a |
 | cw\_alarms\_terraformer | github.com/cisagov/instance-cw-alarms-tf-module | n/a |
+| cw\_alarms\_windows | github.com/cisagov/instance-cw-alarms-tf-module | n/a |
 | email\_sending\_domain\_certreadrole | github.com/cisagov/cert-read-role-tf-module | n/a |
 | guacamole\_certreadrole | github.com/cisagov/cert-read-role-tf-module | n/a |
 | read\_terraform\_state | github.com/cisagov/terraform-state-read-role-tf-module | n/a |
--- a/README.md
+++ b/README.md
@@ -94,6 +94,16 @@ the COOL environment.
 
 | Name | Source | Version |
 |------|--------|---------|
+| cw\_alarms\_assessor\_portal | github.com/cisagov/instance-cw-alarms-tf-module | first-commits |
+| cw\_alarms\_debiandesktop | github.com/cisagov/instance-cw-alarms-tf-module | first-commits |
+| cw\_alarms\_gophish | github.com/cisagov/instance-cw-alarms-tf-module | first-commits |
+| cw\_alarms\_guacamole | github.com/cisagov/instance-cw-alarms-tf-module | first-commits |
+| cw\_alarms\_kali | github.com/cisagov/instance-cw-alarms-tf-module | first-commits |
+| cw\_alarms\_nessus | github.com/cisagov/instance-cw-alarms-tf-module | first-commits |
+| cw\_alarms\_pentestportal | github.com/cisagov/instance-cw-alarms-tf-module | first-commits |
+| cw\_alarms\_samba | github.com/cisagov/instance-cw-alarms-tf-module | first-commits |
+| cw\_alarms\_teamserver | github.com/cisagov/instance-cw-alarms-tf-module | first-commits |
+| cw\_alarms\_terraformer | github.com/cisagov/instance-cw-alarms-tf-module | first-commits |
 | email\_sending\_domain\_certreadrole | github.com/cisagov/cert-read-role-tf-module | n/a |
 | guacamole\_certreadrole | github.com/cisagov/cert-read-role-tf-module | n/a |
 | read\_terraform\_state | github.com/cisagov/terraform-state-read-role-tf-module | n/a |
--- a/README.md
+++ b/README.md
@@ -94,16 +94,16 @@ the COOL environment.
 
 | Name | Source | Version |
 |------|--------|---------|
-| cw\_alarms\_assessor\_portal | github.com/cisagov/instance-cw-alarms-tf-module | first-commits |
-| cw\_alarms\_debiandesktop | github.com/cisagov/instance-cw-alarms-tf-module | first-commits |
-| cw\_alarms\_gophish | github.com/cisagov/instance-cw-alarms-tf-module | first-commits |
-| cw\_alarms\_guacamole | github.com/cisagov/instance-cw-alarms-tf-module | first-commits |
-| cw\_alarms\_kali | github.com/cisagov/instance-cw-alarms-tf-module | first-commits |
-| cw\_alarms\_nessus | github.com/cisagov/instance-cw-alarms-tf-module | first-commits |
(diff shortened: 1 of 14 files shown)


dav3r (approved):
👍 👍

mcdonnnj (changes requested):
This makes sense. I'm only flagging this for changes because I do not see Windows instances getting CloudWatch alarms. They have the CloudWatch Agent (with a default configuration) installed so at least *some* of these alarms should function. Is there a particular reason they were not included?

jsf9k:
> This makes sense. I'm only flagging this for changes because I do not see Windows instances getting CloudWatch alarms. They have the CloudWatch Agent (with a default configuration) installed so at least _some_ of these alarms should function. Is there a particular reason they were not included?

I honestly didn't think about the Windows instances.  I suppose the `AWS/EC2` metrics are the same for the Windows instances, but [my understanding is that the `CWAgent` metrics are completely different for Windows instances](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/metrics-collected-by-CloudWatch-agent.html).

Thoughts:
- We should probably consider shipping a custom AWS CloudWatch Agent config file with the Windows AMIs.
- For now, I could modify [cisagov/instance-cw-alarms-tf-module](https://github.com/cisagov/instance-cw-alarms-tf-module) to allow for a `windows` input variable* that, if set to `true`, signals that a minimal set of alarms using only metrics from the `AWS/EC2` namespace should be created.**

How would you like to proceed?

\* Perhaps `no_cwagent` is a better name for the input variable, although not technically true in the case of Windows.  I can see such a flag being useful for adding alarms for instances that do not have the AWS CloudWatch Agent installed or enabled.
\*\* Should this include the CPU utilization alarm, or only the system status check alarms?

jsf9k:
> > This makes sense. I'm only flagging this for changes because I do not see Windows instances getting CloudWatch alarms. They have the CloudWatch Agent (with a default configuration) installed so at least _some_ of these alarms should function. Is there a particular reason they were not included?
> 
> I honestly didn't think about the Windows instances. I suppose the `AWS/EC2` metrics are the same for the Windows instances, but [my understanding is that the `CWAgent` metrics are completely different for Windows instances](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/metrics-collected-by-CloudWatch-agent.html).
> 
> Thoughts:
> 
> * We should probably consider shipping a custom AWS CloudWatch Agent config file with the Windows AMIs.
> * For now, I could modify [cisagov/instance-cw-alarms-tf-module](https://github.com/cisagov/instance-cw-alarms-tf-module) to allow for a `windows` input variable* that, if set to `true`, signals that a minimal set of alarms using only metrics from the `AWS/EC2` namespace should be created.**
> 
> How would you like to proceed?
> 
> * Perhaps `no_cwagent` is a better name for the input variable, although not technically true in the case of Windows. I can see such a flag being useful for adding alarms for instances that do not have the AWS CloudWatch Agent installed or enabled. ** Should this include the CPU utilization alarm, or only the system status check alarms?

In the absence of input, I will be proceeding with adding a `no_cloudwatch_agent` variable for the [cisagov/instance-cw-alarms-tf-module](https://github.com/cisagov/instance-cw-alarms-tf-module) Terraform module.  This will prevent any alarms based on CloudWatch metrics written by the CloudWatch Agent from being created.  This seems like something that could be generally useful, and it will handle this case as well.

Should we later establish a standard set of alarms based on Windows CloudWatch metrics, I will happily add an additional `is_windows` variable.

jsf9k:
> > > This makes sense. I'm only flagging this for changes because I do not see Windows instances getting CloudWatch alarms. They have the CloudWatch Agent (with a default configuration) installed so at least _some_ of these alarms should function. Is there a particular reason they were not included?
> > 
> > 
> > I honestly didn't think about the Windows instances. I suppose the `AWS/EC2` metrics are the same for the Windows instances, but [my understanding is that the `CWAgent` metrics are completely different for Windows instances](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/metrics-collected-by-CloudWatch-agent.html).
> > Thoughts:
> > 
> > * We should probably consider shipping a custom AWS CloudWatch Agent config file with the Windows AMIs.
> > * For now, I could modify [cisagov/instance-cw-alarms-tf-module](https://github.com/cisagov/instance-cw-alarms-tf-module) to allow for a `windows` input variable* that, if set to `true`, signals that a minimal set of alarms using only metrics from the `AWS/EC2` namespace should be created.**
> > 
> > How would you like to proceed?
> > 
> > * Perhaps `no_cwagent` is a better name for the input variable, although not technically true in the case of Windows. I can see such a flag being useful for adding alarms for instances that do not have the AWS CloudWatch Agent installed or enabled. ** Should this include the CPU utilization alarm, or only the system status check alarms?
> 
> In the absence of input, I will be proceeding with adding a `no_cloudwatch_agent` variable for the [cisagov/instance-cw-alarms-tf-module](https://github.com/cisagov/instance-cw-alarms-tf-module) Terraform module. This will prevent any alarms based on CloudWatch metrics written by the CloudWatch Agent from being created. This seems like something that could be generally useful, and it will handle this case as well.
> 
> Should we later establish a standard set of alarms based on Windows CloudWatch metrics, I will happily add an additional `is_windows` variable.

Please see cisagov/instance-cw-alarms-tf-module#13 and commit c8a1ccf, and note that the GitHub Actions checks will pass once cisagov/instance-cw-alarms-tf-module#13 is merged.

README.md
```diff
@@ -94,6 +94,16 @@ the COOL environment.
...
+| cw\_alarms\_assessor\_portal | github.com/cisagov/instance-cw-alarms-tf-module | n/a |
+| cw\_alarms\_debiandesktop | github.com/cisagov/instance-cw-alarms-tf-module | n/a |
+| cw\_alarms\_gophish | github.com/cisagov/instance-cw-alarms-tf-module | n/a |
+| cw\_alarms\_guacamole | github.com/cisagov/instance-cw-alarms-tf-module | n/a |
+| cw\_alarms\_kali | github.com/cisagov/instance-cw-alarms-tf-module | n/a |
+| cw\_alarms\_nessus | github.com/cisagov/instance-cw-alarms-tf-module | n/a |
+| cw\_alarms\_pentestportal | github.com/cisagov/instance-cw-alarms-tf-module | n/a |
+| cw\_alarms\_samba | github.com/cisagov/instance-cw-alarms-tf-module | n/a |
+| cw\_alarms\_teamserver | github.com/cisagov/instance-cw-alarms-tf-module | n/a |
+| cw\_alarms\_terraformer | github.com/cisagov/instance-cw-alarms-tf-module | n/a |
```
mcdonnnj, line 106: Would you re-generate from `terraform-docs` to pick up the `cw_alarms_windows` module?

mcdonnnj (approved):
Thanks for getting something in place for Windows instances (sorry for forgetting to comment on here with input about a path forward). I only had a comment about updating the `terraform-docs` output in the README.

README.md
jsf9k: Please see commit 934569a.

PR #192, opened 2022-05-25
Fix bug in validation equality check

jsf9k:
## 🗣 Description ##

This pull request fixes a bug in an equality check in a Terraform variable's validation code.

This is necessary because the `[]` operator in Terraform actually creates a tuple, and a tuple will always compare to `false` against a list because they are different types.  [The Terraform documentation](https://www.terraform.io/language/expressions/operators#equality-operators) warns against this very situation.

I noticed this flaw yesterday when helping @mkreckel with some issues he was having creating COOL assessment environments.  At that time I temporarily commented out the validation code as a workaround.

## 🧪 Testing ##

I used the `terraform console` command to verify that this change indeed fixes the issue:
```terraform
$ terraform console -var-file=envXX-production.tfvars
<snip>

> var.email_sending_domains
tolist([
  "abc.org",
  "def.com",
])
> var.email_sending_domains == [for d in var.email_sending_domains : lower(d)]
false
> var.email_sending_domains == tolist([for d in var.email_sending_domains : lower(d)])
true

<snip>
```

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -97,7 +97,14 @@ variable "email_sending_domains" {
   default     = ["example.com"]
 
   validation {
-    condition     = var.email_sending_domains == [for d in var.email_sending_domains : lower(d)]
+    # Note that [] actually creates a tuple, which will always compare
+    # to false against a list because a list and a tuple are different
+    # types.  Therefore the tolist() on the right-hand side is
+    # necessary.
+    #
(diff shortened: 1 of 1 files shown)


dav3r (approved):
Nice comment to go along with this fix 👍

mcdonnnj (approved):
🕵️‍♂️

PR #197, opened 2022-06-16
Inject a dependency of the session manager module on the corresponding policy

jsf9k:
## 🗣 Description ##

This pull request copies a variable together with its default value over from the [cisagov/session-manager-tf-module](https://github.com/cisagov/session-manager-tf-module) so that it its value can be specified without referring to the module.  This allows us to impose a dependency of the module on the policy attachment that allows for the creation of its resources; otherwise, this dependency will be cyclical.

This change gets rid of the dangling CloudWatch log group and SSM document resources that can result when the policy that allows for the creation and destruction of these resources is deleted before the resources are destroyed.

## 🧪 Testing ##

I created and destroyed `env6-staging` several times without issue using these changes.

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -220,8 +220,8 @@ variable "read_terraform_state_role_name" {
 }
 
 # This variable is copied over from cisagov/session-manager-tf-module
-# so that it its value can be specified outside of that module.  This
-# allows is to impose a dependency of the module on the policy that
+# so that its value can be specified outside of that module.  This
+# allows us to impose a dependency of the module on the policy that
 # allows for the creation of its resources; otherwise, this dependency
 # will be cyclical.
 variable "session_cloudwatch_log_group_name" {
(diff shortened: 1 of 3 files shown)


variables.tf
```diff
@@ -219,6 +219,17 @@ variable "read_terraform_state_role_name" {
   default     = "ReadCoolAssessmentTerraformTerraformState-%s"
 }
 
+# This variable is copied over from cisagov/session-manager-tf-module
+# so that it its value can be specified outside of that module.  This
+# allows is to impose a dependency of the module on the policy that
+# allows for the creation of its resources; otherwise, this dependency
+# will be cyclical.
```
dav3r: Typos:
```suggestion
# This variable is copied over from cisagov/session-manager-tf-module
# so that its value can be specified outside of that module.  This
# allows us to impose a dependency of the module on the policy that
# allows for the creation of its resources; otherwise, this dependency
# will be cyclical.
```

dav3r (approved):
Good fix. 👍 

Please note the two small typos I mentioned.

mcdonnnj (approved):
LGTM ✔

PR #205, opened 2022-11-30
Add additional permissions to the Terraformer policy

dav3r:
#  #

## 🗣 Description ##

This PR adds additional permissions to the Terraformer policy related to routing tables, VPC peering connections, and the S3 gateway endpoint.

## 💭 Motivation and context ##

These permissions were requested by the Advanced Operations team so that they could further customize the infrastructure that they build out for their assessments.

This PR is related to these others:
* https://github.com/cisagov/cool-assessment-terraform/pull/206
* https://github.com/cisagov/cool-sharedservices-openvpn/pull/63

## 🧪 Testing ##

I verified that these changes could be successfully applied in a Staging environment.

These policy changes were tested by @quietpirate in Staging, who verified that he was able to successfully deploy and destroy the resources needed by the AO team.

## ✅ Pre-approval checklist ##

- [x] This PR has an informative and human-readable title.
- [x] Changes are limited to a single goal - *eschew scope creep!*
- [x] All relevant type-of-change labels have been added.
- [x] I have read the [CONTRIBUTING](../blob/develop/CONTRIBUTING.md) document.
- [x] These code changes follow [cisagov code standards](https://github.com/cisagov/development-guide).
- [x] All new and existing tests pass.

PR #206, opened 2022-12-02
Allow additional ports for AO VPN endpoints

dav3r:
#  #

## 🗣 Description ##

This PR opens up some additional ports in the network ACLs for the private subnets.

These additional ports will be used to communicate with various VPN endpoints created and used by the Advanced Operations team in their assessments.

This PR is related to these others:
* https://github.com/cisagov/cool-assessment-terraform/pull/205
* https://github.com/cisagov/cool-sharedservices-openvpn/pull/63

## 🧪 Testing ##

I verified that these changes could be successfully applied in a Staging environment.

These networking changes were tested by @quietpirate in Staging, who verified that he was able to successfully communicate with an instance in an assessment environment via the newly-opened ports.


private_acl_rules.tf
```diff
@@ -214,6 +231,21 @@ resource "aws_network_acl_rule" "private_ingress_from_operations_via_https" {
   to_port        = 443
 }
 
+# Allow ingress from private IPs via all ports and protocols
```
jsf9k: I suggest adding a bit of text here explaining why `172.16.0.0/12` is the only private subnet we care about.  Is that what all the assessors are using for the VMs they spin up on their MacBooks?

Also, these `172.16.0.0/12` private IPs are different than the private IPs in the private subnets in the VPC, but I feel like we're calling them both "private" and that is confusing.  I suggest we find another name for these `172.16.0.0/12` private subnets to reduce confusion.

private_acl_rules.tf
```diff
@@ -54,6 +54,23 @@ resource "aws_network_acl_rule" "private_ingress_from_operations_mattermost_web"
   from_port      = 8065
   to_port        = 8065
 }
+# Allow ingress from first private subnet via UDP ephemeral ports.
```
jsf9k: Is this allowing ingress from the first private subnet _to_ the private subnets?  I'm confused as to why this would be necessary.

private_acl_rules.tf
```diff
@@ -312,6 +344,23 @@ resource "aws_network_acl_rule" "private_egress_to_cool_via_ephemeral_ports" {
   to_port        = 65535
 }
 
+# Allow egress to COOL Shared Services via UDP ephemeral ports
+#
+# For: Assessment team access to Advanced Operations VPN endpoints
```
jsf9k, line 355: If I'm not mistaken, this is the bit that was confusing us earlier this week.  Did we ever find any documentation that really explained why this is necessary?  Can we include some explanatory text and/or a link here?

private_acl_rules.tf
```diff
@@ -380,3 +429,18 @@ resource "aws_network_acl_rule" "private_egress_to_cool_via_ipa_ports" {
   from_port      = each.value.port
   to_port        = each.value.port
 }
+
+# Allow egress to private IPs via all ports and protocols
```
jsf9k: I suggest adding a bit of text here explaining why `172.16.0.0/12` is the only private subnet we care about.  Is that what all the assessors are using for the VMs they spin up on their MacBooks?

Also, these `172.16.0.0/12` private IPs are different than the private IPs in the private subnets in the VPC, but I feel like we're calling them both "private" and that is confusing.  I suggest we find another name for these `172.16.0.0/12` private subnets to reduce confusion.

jsf9k (changes requested):
I'm requesting that some comments be expanded upon, and possibly some nomenclature changed, as I am still confused.

private_acl_rules.tf
```diff
@@ -214,6 +231,21 @@ resource "aws_network_acl_rule" "private_ingress_from_operations_via_https" {
...
+resource "aws_network_acl_rule" "private_ingress_from_private_ips_via_all_ports" {
+  provider = aws.provisionassessment
+  for_each = toset(var.private_subnet_cidr_blocks)
+
+  network_acl_id = aws_network_acl.private[each.value].id
+  egress         = false
+  protocol       = "all"
+  rule_number    = 210 + index(var.private_subnet_cidr_blocks, each.value)
+  rule_action    = "allow"
+  cidr_block     = "172.16.0.0/12"
```
mcdonnnj, line 246: Are we OK hard coding this value?

private_acl_rules.tf
```diff
@@ -380,3 +429,18 @@ resource "aws_network_acl_rule" "private_egress_to_cool_via_ipa_ports" {
...
+resource "aws_network_acl_rule" "private_egress_to_private_ips_via_all_ports" {
+  provider = aws.provisionassessment
+  for_each = toset(var.private_subnet_cidr_blocks)
+
+  network_acl_id = aws_network_acl.private[each.value].id
+  egress         = true
+  protocol       = "all"
+  rule_number    = 410 + index(var.private_subnet_cidr_blocks, each.value)
+  rule_action    = "allow"
+  cidr_block     = "172.16.0.0/12"
```
mcdonnnj, line 445: Are we OK hard coding this value?

mcdonnnj (approved):
Seems reasonable enough once Shane's concerns are addressed.

private_acl_rules.tf
dav3r: You bring up an excellent point.  Please take a look at https://github.com/cisagov/cool-assessment-terraform/pull/206/commits/0da00676cce3ca3b212ea163db0f1df9a268ea8e and tell me if that works for you.

private_acl_rules.tf
dav3r: Please take a look at https://github.com/cisagov/cool-assessment-terraform/pull/206/commits/0da00676cce3ca3b212ea163db0f1df9a268ea8e and tell me if that works for you.

private_acl_rules.tf
dav3r: You are correct- I had this rule somewhat backwards before.  I have corrected it, improved the comment (hopefully), and tested it to my satisfaction with https://github.com/cisagov/cool-assessment-terraform/pull/206/commits/34051abde3b014b5e82eda1945769a0d7edabbb3.  Let me know what you think.

private_acl_rules.tf
jsf9k: I'd still like a comment to explicitly state that the `172.16.0.0/12` network is where the assessors are running the VMs on their Macbooks.  "Local VM" could be local to the user's machine or local to the AWS instance and hence is ambiguous, and I won't remember what it means when I revisit this code.

Also, for this to work we require the network on the end user's machine to be set up in a particular way.  (This may be the default way for VMWare on OSX, but on my machine, e.g.,  it would likely be trying to route packets to Docker containers.)  This is a big deal that we shouldn't lose sight of.

For these reasons, I think it makes sense to write a few sentences explaining exactly what is going on here.

private_acl_rules.tf
dav3r: I think you might be mixing this up with the changes I made in https://github.com/cisagov/cool-assessment-terraform/pull/206/commits/a3f8be86353918eaf593eeda6a3b45223c08c3cb and then just corrected in https://github.com/cisagov/cool-assessment-terraform/pull/206/commits/34051abde3b014b5e82eda1945769a0d7edabbb3.  This rule is necessary to allow return traffic on UDP ephemeral ports to flow from the transit gateway attachment to the Shared Services subnet (e.g. VPN-connected laptops used by assessors).

I think I could make this rule more restrictive so that it only allows this traffic out from the first private subnet (where the TGW attachment lives), but since we already [allow TCP ephemeral ports back to COOL Shared Services from both private subnets](https://github.com/cisagov/cool-assessment-terraform/blob/develop/private_acl_rules.tf#L298-L313), I figured it was fine to do the same for UDP.

jsf9k: OK, I got it.

private_acl_rules.tf
dav3r: Got it- how about something like this?

```
# Allow ingress from 172.16.0.0/12 via all ports and protocols.  172.16.0.0/12 is the private
# network block that has been chosen by the Advanced Operations team where they will
# run their mission-related virtual machines on their VPN-connected laptops.
# 
# For: Advanced Operations communications with virtual machines running on an operator's
# VPN-connected laptop
```

jsf9k: Yes, that scratches the itch.  Thank you.

dav3r: No, thank you!  Updated in https://github.com/cisagov/cool-assessment-terraform/pull/206/commits/84cb2816feecff60f1eae8d7557755349f19923e.

private_acl_rules.tf
dav3r: Please see https://github.com/cisagov/cool-assessment-terraform/pull/206/commits/84cb2816feecff60f1eae8d7557755349f19923e for latest verbiage.

private_acl_rules.tf
jsf9k: I see commit 84cb281.  Thank you.

jsf9k (approved):

PR #208, opened 2023-01-20
Support Egress-Assess Instances

m1j09830:
## 🗣 Description ##

This PR contains the code to support deploying Egress-Assess instances to the operational environments.

The Egress-Assess AMI created by this [PR](https://github.com/cisagov/egress-assess-packer/pull/11), needs to be supported by the cool-assessment-terraform code so that it can be deployed to the operational environments. 

## 🧪 Testing ##

Deployed the Egress-Assess instance to a staging environment using the updated terraform code where it was successfully tested for proper functionality.

## ✅ Post-merge checklist ##

- [x] Educate NOM regarding new egress-assess instance type


dav3r (approved):
Strong work @m1j09830 - you must've had an incredible mentor! 🥇

operations_acl_rules.tf
```diff
@@ -145,7 +145,7 @@ resource "aws_network_acl_rule" "operations_ingress_from_anywhere_via_ports_1024
   network_acl_id = aws_network_acl.operations.id
   egress         = false
   protocol       = each.value
-  rule_number    = 160 + index(local.tcp_and_udp, each.value)
+  rule_number    = 170 + index(local.tcp_and_udp, each.value)
```
jsf9k, line 148: Why did this rule number (and the one just below this) need to change?

jsf9k (commented):
This looks good, but I have one question.

operations_acl_rules.tf
dav3r: Good eye- this was done because in the course of testing the Egress Assess instance, the previous NACL rule group (`aws_network_acl_rule.operations_ingress_from_anywhere_via_allowed_ports`) required more than 10 NACL rules (as specified by `var.inbound_ports_allowed["egressassess"]` in our test `tfvars` file) and that broke our old rule numbering scheme.  

To remedy this, we bumped up the subsequent rule numbers by 10 to avoid any conflicts.  We are now able to support use cases of up to 20 `operations_ingress_from_anywhere_via_allowed_ports` NACL rules instead of 10.

jsf9k: Makes sense.

jsf9k (approved):

egressassess_sg.tf
```diff
@@ -0,0 +1,55 @@
...
+  vpc_id = aws_vpc.assessment.id
+
+  tags = {
+    "Name" = "Egress-Assess"
+  }
+}
+
+# Allow egress to anywhere via HTTP and HTTPS
+#
+# For: Package downloads and updates
```
mcdonnnj, line 14: Is this egress also for the [Auto-Egress-Assess](https://github.com/tarrell13/Auto-Egress-Assess) tool? If so we should explicitly mention it here. Otherwise should there be other egress rules for that tool's functionality?

mcdonnnj (changes requested):

egressassess_sg.tf
m1j09830: This instance only needs to be able to download packages and and perform updates. There are no other egress needs for this tool. The comment as is accurately represents the needs for the instance and tool.

jsf9k: This egress is for the Auto-Egress-Assess tool.  Auto-Egress-Assess is the new name of their version of the Egress-Assess tool.  They "worked around" the issue of merging back into the upstream by changing their fork into a new repo.

mcdonnnj: @m1j09830 I'm not really understanding your response. If this rule were to be removed would the [Auto-Egress-Assess](https://github.com/tarrell13/Auto-Egress-Assess) tool still function?

dav3r: Yes, the tool would still function- it does not require any egress ports.  80 and 443 are only used for pulling in packages. 

The ports that the tool listens on are opened via `var.inbound_ports_allowed`:
https://github.com/cisagov/cool-assessment-terraform/blob/2dd006e6228253e22e802393c749876bd080c1b1/variables.tf#L136-L164

m1j09830: @mcdonnnj does @dav3r's response make more sense? I apologize if I'm not being clear but he covered what I was trying to say much more eloquently than I could.

mcdonnnj: Went and RTFM of the tool so I have a better understanding of what it does. Assessors are (I'm assuming) only using the server portion of the code on this instance and thus only need to set ingress rules for the tool to function.

mcdonnnj (approved):
LGTM ✔

PR #210, opened 2023-02-22
Add variables for assessment ID and type

dav3r:
#  #

## 🗣 Description ##

This PR adds new optional variables for assessment identifier, assessment type, valid assessment identifier regular expression, and valid assessment types.  If provided, the assessment identifier is validated against the valid assessment identifier regular expression and the assessment type is validated against the list of valid assessment types. 

## 💭 Motivation and context ##

Defining assessment types and identifiers is a necessary precursor for the work planned for https://github.com/cisagov/cool-system-internal/issues/3.  A subsequent PR will add a script that can copy assessment results to an S3 bucket using a standard naming scheme based on assessment type and identifier.

## 🧪 Testing ##

I verified the following:
* When none of these (optional) new variables are supplied, no validation errors are thrown during a `terraform plan` or `terraform apply`.
* When an assessment ID is provided, but no valid ID regex is provided, an error like the following is thrown:
```
│ Error: Resource precondition failed
│ 
│   on assessment_validation.tf line 15, in resource "null_resource" "validate_assessment_id":
│   15:       condition     = length(regexall(var.valid_assessment_id_regex, var.assessment_id)) == 1
│     ├────────────────
│     │ var.assessment_id is "ASMT1234"
│     │ var.valid_assessment_id_regex is ""
│ 
│ Invalid assessment identifier provided: ASMT1234.  Valid assessment identifiers must match this regular expression:
```
* When an assessment ID is provided that does not match the provided regex for valid IDs, an error like the following is thrown:
```
│ Error: Resource precondition failed
│ 
│   on assessment_validation.tf line 15, in resource "null_resource" "validate_assessment_id":
│   15:       condition     = length(regexall(var.valid_assessment_id_regex, var.assessment_id)) == 1
│     ├────────────────
│     │ var.assessment_id is "ASMT-X23"
│     │ var.valid_assessment_id_regex is "^ASMT[[:digit:]]{4}$"
│ 
│ Invalid assessment identifier provided: ASMT-X23.  Valid assessment identifiers must match this regular expression: ^ASMT[[:digit:]]{4}$
```
* When an assessment ID is provided that matches the provided regex for valid IDs, no validation errors are thrown during a `terraform plan` or `terraform apply`.
* When an assessment type is provided, but no list of valid assessment types is provided, an error like the following is thrown:
```
│ Error: Resource precondition failed
│ 
│   on assessment_validation.tf line 24, in resource "null_resource" "validate_assessment_type":
[...]

PR #207, opened 2022-12-20
Add ability to write findings data to bucket

dav3r:
#  #

## 🗣 Description ##

This PR adds a script (and related resources) that allows an assessor to copy data to the assessment findings bucket.

## 💭 Motivation and context ##

We want to allow data to be copied from COOL assessment environments to the assessment findings bucket.

Related to:
* https://github.com/cisagov/findings-data-import-terraform/pull/34
* https://github.com/cisagov/cool-accounts/pull/132

This is part of the work required for https://github.com/cisagov/cool-system-internal/issues/122.

## 🧪 Testing ##

@jsf9k and I successfully applied this Terraform in a Staging environment (after I applied the Terraform in https://github.com/cisagov/findings-data-import-terraform/pull/34 and https://github.com/cisagov/cool-accounts/pull/132) and we confirmed the following:
* The new `copy-findings-data-to-bucket.sh` script was correctly generated by cloud-init.
* We were able to successfully copy a JSON file to the assessment findings bucket using `copy-findings-data-to-bucket.sh`.
* For assessments that do not require the ability to copy data to the findings bucket, the `findings_data_bucket_name` Terraform variable can be set to an empty string (`""`).

## ✅ Pre-approval checklist ##

- [x] This PR has an informative and human-readable title.
- [x] Changes are limited to a single goal - *eschew scope creep!*
- [x] All relevant type-of-change labels have been added.
- [x] I have read the [CONTRIBUTING](../blob/develop/CONTRIBUTING.md) document.
- [x] These code changes follow [cisagov code standards](https://github.com/cisagov/development-guide).
- [x] All relevant repo and/or project documentation has been updated
      to reflect the changes in this PR.
- [x] All new and existing tests pass.

## ✅ Post-merge checklist ##

- [x] Coordinate with NOM to ensure they know how to update Production tfvars files.

PR #213, opened 2023-03-06
Support assessment artifact export

dav3r:
#  #

## 🗣 Description ##

This PR adds optional support for archiving and then exporting assessment artifacts to an S3 bucket.

This PR adds on to the work done in #210 and resolves https://github.com/cisagov/cool-system-internal/issues/3.

## 🧪 Testing ##

I confirmed that the new validation was working as intended for `var.assessment_artifact_export_map` by providing an assessment type that was not included in the list of valid types:
```
│ Error: Resource precondition failed
│ 
│   on assessment_validation.tf line 15, in resource "null_resource" "validate_assessment_artifact_export_map":
│   15:       condition     = length([for k in keys(var.assessment_artifact_export_map) : k if !contains(var.valid_assessment_types, k)]) == 0
│     ├────────────────
│     │ var.assessment_artifact_export_map is map of string with 4 elements
│     │ var.valid_assessment_types is list of string with 4 elements
│ 
│ Invalid assessment type(s) provided in assessment_artifact_export_map: WiFi.  Valid types are: PenTest, Phishing, RedTeam
```

Next, I configured all of the variables related to assessment export and ran `terraform apply` to deploy a Kali instance into a test environment.  I confirmed that the `artifact_export_credentials` and `archive-artifact-data-to-bucket.sh` files were both created and looked correct.

Finally, I ran `archive-artifact-data-to-bucket.sh` successfully (i.e. with a valid parameter) and unsuccessfully (without providing the required parameter).  In both cases, the script performed as intended, throwing the correct error message in the latter case.

## ✅ Post-merge checklist ##
- [x] Populate Production SSM Parameter Store data


variables.tf
```diff
@@ -48,6 +48,18 @@ variable "assessmentfindingsbucketwrite_sharedservices_policy_name" {
...
+  type        = bool
+  description = "Whether or not to enable the export of assessment artifacts to an S3 bucket.  If this is set to true, then the following variables should also be configured appropriately: assessment_artifact_export_map, ssm_key_artifact_export_access_key_id, ssm_key_artifact_export_secret_access_key, ssm_key_artifact_export_bucket_name, and ssm_key_artifact_export_bucket_region."
+  default     = false
+}
+
+variable "assessment_artifact_export_map" {
+  type        = map(string)
+  description = "A map whose keys are assessment types and whose values are the prefixes for what an assessment artifact will be named when it is exported to the S3 bucket contained in the SSM parameter specified by the ssm_key_artifact_export_bucket_name variable (e.g. { \"PenTest\" : \"pentest/PT\", \"Phishing\" : \"phishing/PH\", \"RedTeam\" : \"redteam/RT\" }). Note that prefixes can include a path within the bucket.  For example, if the prefix is \"pentest/PT\" and the assessment ID is \"ASMT1234\", then the corresponding artifact will be exported to \"bucket-name/pentest/PT-ASMT1234\" when the copy-artifact-data-to-bucket.sh script is run."
```
mcdonnnj: Just aligning the object name with what the script is creating.

```suggestion
  description = "A map whose keys are assessment types and whose values are the prefixes for what an assessment artifact will be named when it is exported to the S3 bucket contained in the SSM parameter specified by the ssm_key_artifact_export_bucket_name variable (e.g. { \"PenTest\" : \"pentest/PT\", \"Phishing\" : \"phishing/PH\", \"RedTeam\" : \"redteam/RT\" }). Note that prefixes can include a path within the bucket.  For example, if the prefix is \"pentest/PT\" and the assessment ID is \"ASMT1234\", then the corresponding artifact will be exported to \"bucket-name/pentest/PT-ASMT1234.tgz\" when the copy-artifact-data-to-bucket.sh script is run."
```

README.md
```diff
@@ -475,6 +480,8 @@ the COOL environment.
...
+| assessment\_artifact\_export\_enabled | Whether or not to enable the export of assessment artifacts to an S3 bucket.  If this is set to true, then the following variables should also be configured appropriately: assessment\_artifact\_export\_map, ssm\_key\_artifact\_export\_access\_key\_id, ssm\_key\_artifact\_export\_secret\_access\_key, ssm\_key\_artifact\_export\_bucket\_name, and ssm\_key\_artifact\_export\_bucket\_region. | `bool` | `false` | no |
+| assessment\_artifact\_export\_map | A map whose keys are assessment types and whose values are the prefixes for what an assessment artifact will be named when it is exported to the S3 bucket contained in the SSM parameter specified by the ssm\_key\_artifact\_export\_bucket\_name variable (e.g. { "PenTest" : "pentest/PT", "Phishing" : "phishing/PH", "RedTeam" : "redteam/RT" }). Note that prefixes can include a path within the bucket.  For example, if the prefix is "pentest/PT" and the assessment ID is "ASMT1234", then the corresponding artifact will be exported to "bucket-name/pentest/PT-ASMT1234" when the copy-artifact-data-to-bucket.sh script is run. | `map(string)` | `{}` | no |
```
mcdonnnj: Corresponding change to the README.

```suggestion
| assessment\_artifact\_export\_map | A map whose keys are assessment types and whose values are the prefixes for what an assessment artifact will be named when it is exported to the S3 bucket contained in the SSM parameter specified by the ssm\_key\_artifact\_export\_bucket\_name variable (e.g. { "PenTest" : "pentest/PT", "Phishing" : "phishing/PH", "RedTeam" : "redteam/RT" }). Note that prefixes can include a path within the bucket.  For example, if the prefix is "pentest/PT" and the assessment ID is "ASMT1234", then the corresponding artifact will be exported to "bucket-name/pentest/PT-ASMT1234.tgz" when the copy-artifact-data-to-bucket.sh script is run. | `map(string)` | `{}` | no |
```

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
```diff
@@ -0,0 +1,74 @@
...
+# Write the script.  Note that we wrap the delimited in quotes to
+# prevent shell variable substitution.
+cat > "$path" << "EOF"
+#!/usr/bin/env bash
+
+# This script creates a gzipped tar archive of the directory containing
+# assessment artifacts and then copies that archive to the appropriate S3
+# bucket.
+#
+# Usage:  copy-artifact-data-to-bucket.sh /path/to/artifacts_directory
```
mcdonnnj: Aligning with the path being written to by the cloudinit script.

```suggestion
# Usage: archive-artifact-data-to-bucket.sh /path/to/artifacts_directory
```

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
```diff
@@ -0,0 +1,74 @@
...
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+if [ $# -ne 1 ]; then
+  echo "Usage:  copy-artifact-data-to-bucket.sh /path/to/artifacts_directory"
```
mcdonnnj: Here as well.

```suggestion
  echo "Usage: archive-artifact-data-to-bucket.sh /path/to/artifacts_directory"
```

kali_cloud_init.tf
```diff
@@ -121,4 +121,64 @@ data "cloudinit_config" "kali_cloud_init_tasks" {
...
+  #
+  # Input variables are:
+  # * aws_access_key_id - the AWS access key ID
+  # * aws_region - the AWS region of the S3 bucket
+  # * aws_secret_access_key - the AWS secret access key
+  # * permissions - the octal permissions to assign the AWS configuration
+  # * vnc_username - the username associated with the VNC user
+  dynamic "part" {
+    # Only include this block if var.assessment_artifact_export_enabled is true.
+    for_each = var.assessment_artifact_export_enabled ? [0] : []
```
mcdonnnj, line 140: We're using `for_each` because it's a `dynamic` block right?

variables.tf
```diff
@@ -264,6 +276,30 @@ variable "session_cloudwatch_log_group_name" {
...
+
+variable "ssm_key_artifact_export_bucket_name" {
+  type        = string
+  description = "The AWS SSM Parameter Store parameter that contains the name of the assessment artifact export bucket (e.g. \"/assessment_artifact_export/bucket\")."
+  default     = "/assessment_artifact_export/bucket"
+}
+
+variable "ssm_key_artifact_export_bucket_region" {
+  type        = string
+  description = "The AWS SSM Parameter Store parameter that contains the region of the assessment artifact export bucket (e.g. \"/assessment_artifact_export/region\")."
```
mcdonnnj: I believe this region is for the IAM user that can access the bucket, as buckets themselves are considered global.

mcdonnnj (changes requested):
Strong work 💪💪💪 I have some suggestions and comments for your consideration.

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
+# This is a Terraform template file, and the input variables are
+# passed in via templatefile().
+#
+# shellcheck disable=SC2154
```
felddy, line 5: Do we need a reason here?

cloud-init/write-kali-aws-config-artifact-export.tpl.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+# This is a Terraform template file, and the input variables are
+# passed in via templatefile().
+#
+# shellcheck disable=SC2154
```
felddy, line 5: Do we need a reason here?

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
```diff
@@ -0,0 +1,74 @@
...
+# * permissions - the octal permissions to assign the script
+# * vnc_username - the username associated with the VNC user
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+path=/home/${vnc_username}/archive-artifact-data-to-bucket.sh
+
+# Write the script.  Note that we wrap the delimited in quotes to
```
jsf9k:
```suggestion
# Write the script.  Note that we wrap the delimiter in quotes to
```

kali_cloud_init.tf
dav3r: Correct.  I'm following the same pattern that we established in #207:
https://github.com/cisagov/cool-assessment-terraform/blob/840cbc636c801e9812630eb08c75c89799d00b4d/kali_cloud_init.tf#L78-L81

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
dav3r: We could probably explain this a bit more verbosely, but we use the same text in all of our `*.tpl.sh` files in this directory:
```
# This is a Terraform template file, and the input variables are
# passed in via templatefile().
```

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
```diff
@@ -0,0 +1,74 @@
...
+  echo "Exiting without archiving or copying data to S3."
+  exit 1
+fi
+
+# Change to the parent directory of the artifacts directory
+cd "$(dirname "$1")"
+
+# Create the archive
+echo "Creating archive..."
+tar --create --file ${assessment_id}.tgz --gzip --verbose "$(basename "$1")"
```
jsf9k, line 68: I suggest writing the archive to `/tmp`, unless it is too large to go there.  If it is expected to be huge then I suggest writing it to the network drive.  I don't know how large they are expected to get, but eating up all the root disk would be a bad thing to do.

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
```diff
@@ -0,0 +1,74 @@
...
+
+# Copy the archive to the S3 bucket
+echo "Copying archive to S3..."
+AWS_SHARED_CREDENTIALS_FILE=/home/${vnc_username}/.aws/artifact_export_credentials aws s3 cp ${assessment_id}.tgz "$full_bucket_path"
+
+# Delete the archive
+echo "Deleting archive..."
+rm ${assessment_id}.tgz
```
jsf9k, line 76: If you do change the location where the archive is stored, then you'll need to update this line as well.

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
```diff
@@ -0,0 +1,74 @@
...
+
+# This script is used write out a bash script that can be used to create an
+# assessment artifact archive and copy it to the appropriate S3 bucket.
+
+# Input variables are:
+# * artifact_export_bucket_name - the name of the assessment artifact export S3
+#   bucket
+# * artifact_export_path - the path to copy the artifact to in the S3 bucket
+# * assessment_id - the identifier for the assessment
+# * permissions - the octal permissions to assign the script
```
jsf9k: Is there a reason why these permissions must be specified in octal?  Looking at the script it appears that any format supported by `chmod` should be OK.

cloud-init/write-kali-aws-config-artifact-export.tpl.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# This script is used to write out an AWS configuration to the home
+# directory of a specified user.  The AWS configuration gives the user
+# credentials that allow writing to the assessment artifact export S3 bucket.
+
+# Input variables are:
+# * aws_access_key_id - the AWS access key ID
+# * aws_region - the AWS region of the S3 bucket
+# * aws_secret_access_key - the AWS secret access key
+# * permissions - the octal permissions to assign the AWS
```
jsf9k: Is there a reason why these permissions must be specified in octal?  Looking at the script it appears that any format supported by `chmod` should be OK.

jsf9k (changes requested):
Fixing the one typo is required, but the other suggestions are just that.  Strong work!

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
dav3r: In our use case, the directory to be exported lives on the EFS volume.  Since the script `cd`'s to the parent directory of the artifacts directory, the archive will be created on the EFS volume too.  If that's not good enough, I can modify it so that we use `/tmp` instead.

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
dav3r: I'm following the same pattern that you started in #207:
https://github.com/cisagov/cool-assessment-terraform/blob/840cbc636c801e9812630eb08c75c89799d00b4d/kali_cloud_init.tf#L66-L76

If we change this, we should change it everywhere in this repo.

jsf9k: Unless you see a reason why it has to be octal I'd prefer that you change yours.  I can create an issue for the rest.  (Or you can go ahead and change them here if you're feeling industrious.)

- Geez, if I walked to the middle of the Golden Gate and jumped off would you still follow me?
- Do as I say, not as I do.

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
jsf9k: OK, can you add a comment to that effect above the `tar` command, so I don't get hung up on this point again?

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
dav3r: Fixed in https://github.com/cisagov/cool-assessment-terraform/pull/213/commits/11c4de2f7d51a3c210cc6d7616dce18a55695b30 (along with the source of the typo).

variables.tf
dav3r: Good catch!  I corrected this and some related items in https://github.com/cisagov/cool-assessment-terraform/pull/213/commits/cc26a3ef31d656d09c3ffbf0c845f4fe3b159c6a.

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
dav3r: Comment added in https://github.com/cisagov/cool-assessment-terraform/pull/213/commits/a16d178755267fb27f0ba8ab79fa6d2f90787d7f.

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
dav3r: Are you just looking for me to change the comment verbiage to remove the word "octal"?  Or are you also asking to pass in permission values different than `"0400"` or `"0500"`?

jsf9k: I'm only suggesting changing the description to something like `the permissions to assign the script, specified in either the octal or symbolic formats understood by chmod`.  There is no need to change the actual values passed in.

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
```diff
@@ -0,0 +1,79 @@
...
+# Change to the parent directory of the artifacts directory
+cd "$(dirname "$1")"
+
+# Create the archive
+echo "Creating archive..."
+# Since we are in the parent directory of the artifacts directory, we don't
+# expect to run into any issues related to disk space (in our typical use case,
+# this directory resides on the EFS volume, which should not run out of space).
+# If disk space does become a problem for some reason, this script can be
+# modified to create the archive in a temporary directory.
```
jsf9k: If the EFS volume can't contain the archive, then certainly `/tmp` can't either.
```suggestion
# Since we are in the parent directory of the artifacts directory, we don't
# expect to run into any issues related to disk space.  (In our typical use 
# case, this directory resides on the EFS volume, which is huge.)
```

jsf9k (changes requested):

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
dav3r: Haha, fair point.  I meant that to cover cases where the EFS volume wasn't being used for the artifacts directory and the archive, since there is no requirement for folks to do so.

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
dav3r: Please take a look at https://github.com/cisagov/cool-assessment-terraform/pull/213/commits/674c6efd7485bfdddddc4a82e9b1eba7f40e427c.

cloud-init/write-kali-aws-config-artifact-export.tpl.sh
dav3r: Please take a look at https://github.com/cisagov/cool-assessment-terraform/pull/213/commits/674c6efd7485bfdddddc4a82e9b1eba7f40e427c.

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
jsf9k: Thank you!

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
jsf9k: You also make a good point, now that you have explained it.  How about this?
```suggestion
# Since we are in the parent directory of the artifacts directory, we don't
# expect to run into any issues related to disk space.  (In our typical use 
# case, this directory resides on the EFS volume, which is huge.)
#
# The important thing is to avoid filling up the root disk, so in the event
# that an EFS volume _is not_ being used it probably makes sense to create 
# the archive in /tmp.
```

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
jsf9k: I think this is fine as it stands, although if @felddy wants to suggest new language I will happily consider it.  We already adequately explain the use of `shellcheck disable`.

cloud-init/write-kali-aws-config-artifact-export.tpl.sh
jsf9k: See [here](https://github.com/cisagov/cool-assessment-terraform/pull/213#discussion_r1128306069).

jsf9k (approved):
Thank you for considering my various suggestions.

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
```diff
@@ -0,0 +1,81 @@
...
+
+# This script is used write out a bash script that can be used to create an
+# assessment artifact archive and copy it to the appropriate S3 bucket.
+
+# Input variables are:
+# * artifact_export_bucket_name - the name of the assessment artifact export S3
+#   bucket
+# * artifact_export_path - the path to copy the artifact to in the S3 bucket
+# * assessment_id - the identifier for the assessment
+# * permissions - the octal permissions to assign the script
```
mcdonnnj: Did this get lost along the way?

```suggestion
# * permissions - the permissions to assign the AWS configuration, specified
#   in either the octal or symbolic formats understood by chmod
```

cloud-init/write-kali-aws-config.tpl.sh
```diff
@@ -13,8 +13,8 @@
 # * aws_region - the AWS region where the roles are to be assumed
```
mcdonnnj, line 13: Since this has been dragged in I just want to mention that I think the findings data AWS credentials should be modified to be like the artifact export credentials. Specifically the script name and the name of the generated AWS credentials file. This can be done in a separate PR via an issue. I just want to reach consensus on the suggestion.

variables.tf
```diff
@@ -264,6 +276,30 @@ variable "session_cloudwatch_log_group_name" {
   type        = string
 }
 
+variable "ssm_key_artifact_export_access_key_id" {
+  type        = string
+  description = "The AWS SSM Parameter Store parameter that contains the AWS access key of the user that can write to the assessment artifact export bucket (e.g. \"/assessment_artifact_export/access_key_id\")."
```
mcdonnnj: I'm not sure if it's worth the bother but should these new variables consistently reference `IAM user` (as opposed to just `user`) as is done in the `ssm_key_artifact_export_region` variable?

mcdonnnj (changes requested):
Home stretch but I have a few items for your consideration.

cloud-init/write-kali-aws-config.tpl.sh
dav3r: Yes, I am planning to do that after this PR is merged, but I hadn't mentioned it here yet.  I wasn't planning on changing the script name, but I was definitely going to rename the credentials file.  I intentionally want the names of these scripts to be somewhat different, in order to minimize confusion.

cloud-init/write-archive-artifact-data-to-bucket.tpl.sh
dav3r: Yup - not sure I missed that one and the other one, but thanks!  Cleaned up in https://github.com/cisagov/cool-assessment-terraform/pull/213/commits/df151d554320f677f2ce9e612ff61b716dee8c15.

variables.tf
dav3r: Agreed - no bother at all.  Please see https://github.com/cisagov/cool-assessment-terraform/pull/213/commits/9b2edb863f62cf11d0c94ce0dba98c3f65c16a4d.

mcdonnnj (approved):
LGTM ✔

PR #215, opened 2023-03-31
Add permissions boundary policy to Terraformer role

dav3r:
#  #

## 🗣 Description ##

This PR creates a [permissions boundary policy](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_boundaries.html) and applies it to the Terraformer role.

The rules of the Terraformer boundary policy are:
* Not allowed to modify any resources tagged by the team that deploys the assessment environment (aside from a small number of exceptions that have been granted)
* Not allowed to assume the Guacamole, Samba, and Terraformer instance roles
* Not allowed to modify CloudFormation resources created by Control Tower
* Not allowed to modify or delete the permissions boundary policy
* Not allowed to remove the permissions boundary policy from users and roles that have it
* Not allowed to create users or roles that do not include the permissions boundary
* Not allowed to apply any other permission boundary policies to users or roles

Previously, no IAM permissions were allowed by the Terraformer role because that would've allowed the creation of a policy with admin access.  This change allows the Terraformer role to perform IAM functions (e.g. creating custom roles and policies) as long as they don't break the rules specified in the permissions boundary policy.  This will make the Terraformer role more useful to our users and enable them to deploy a wider variety of resources in their environments.

## 🧪 Testing ##

I deployed these changes to a test environment and attempted to perform activities that should be prohibited by the boundary policy (e.g. modify resources that were tagged by deployment team, replace the boundary policy with a different one, remove the boundary policy from the Terraformer role).  In every case that I tested, I was unable to violate the rules of the boundary policy.

In addition, I confirmed (with help from others) that this boundary policy allows the following:
- [x] Successful deployment of [NuclearPond](https://github.com/DevSecOpsDocs/terraform-nuclear-pond) (with minor changes to the standard NuclearPond Terraform)
- [x] Successful deployment of all Terraform used by the Advanced Operations team

### DIFF
--- a/terraformer_permissionsboundary.tf
+++ b/terraformer_permissionsboundary.tf
@@ -33,8 +33,8 @@ data "aws_iam_policy_document" "terraformer_permissions_boundary_policy_doc" {
     sid = "AllowReadingEC2ResourcesTaggedByTeam"
   }
 
-  # Deny modification of any resources tagged by the team that deploys this
-  # root module, but allow anything else.
+  # Allow modification of any resources, except those tagged by the team that
+  # deploys this root module.
   statement {
     actions = [
       "*",
--- a/terraformer_permissionsboundary.tf
+++ b/terraformer_permissionsboundary.tf
@@ -3,6 +3,7 @@
 # * Deny modification of any resources tagged by the team that deploys this
 #   root module, with some exceptions (detailed below).
 # * Deny use of the Guacamole, Samba, and Terraformer instance roles.
+# * Deny modification of CloudFormation resources created by Control Tower.
 # * Deny modification or deletion of this permissions boundary policy.
 # * Deny removal of this permissions boundary policy from users and roles.
 # * Deny creation of users or roles that do not have this permissions boundary.
@@ -170,7 +171,7 @@ data "aws_iam_policy_document" "terraformer_permissions_boundary_policy_doc" {
   }
 
   # Don't allow Terraformer instances to touch the CloudFormation foo
-  # put in place by ControlTower.  This is not covered by the earlier
+  # put in place by Control Tower.  This is not covered by the earlier
   # statement allowing full access to resources that are not tagged as
   # belonging to the team that deploys this root module, since CloudFormation
   # resources do not accept tags.
--- a/terraformer_permissionsboundary.tf
+++ b/terraformer_permissionsboundary.tf
@@ -190,11 +190,11 @@ data "aws_iam_policy_document" "terraformer_permissions_boundary_policy_doc" {
     sid = "AllowModifyingS3Endpoint"
   }
 
-  # Don't allow Terraformer instances to touch the CloudFormation foo
-  # put in place by Control Tower.  This is not covered by the earlier
-  # statement allowing full access to resources that are not tagged as
-  # belonging to the team that deploys this root module, since CloudFormation
-  # resources do not accept tags.
+  # Don't allow Terraformer instances to touch the CloudFormation foo put in
+  # place by Control Tower.  CloudFormation resources do not accept tags and
+  # thus cannot be tagged with a team tag, which is why this statement is needed
+  # (an earlier statement prevents modification of any resources tagged by the
+  # team that deploys this root module).
   statement {
     actions = [
       "cloudformation:*",
--- a/terraformer_permissionsboundary.tf
+++ b/terraformer_permissionsboundary.tf
@@ -12,6 +12,27 @@
 data "aws_iam_policy_document" "terraformer_permissions_boundary_policy_doc" {
   provider = aws.provisionassessment
 
+  # Allow read-only access to EC2 resources tagged by the team that deploys
+  # this root module.
+  statement {
+    actions = [
+      "ec2:Describe*",
+      "ec2:Get*",
+      "ec2:List*",
+    ]
+    condition {
+      test = "StringEquals"
+      values = [
+        var.tags["Team"],
+      ]
+      variable = "aws:ResourceTag/Team"
+    }
+    resources = [
+      "*",
+    ]
+    sid = "AllowReadingEC2ResourcesTaggedByTeam"
+  }
+
   # Deny modification of any resources tagged by the team that deploys this
   # root module, but allow anything else.
   statement {
@@ -115,7 +136,6 @@ data "aws_iam_policy_document" "terraformer_permissions_boundary_policy_doc" {
       "ec2:CreateSecurityGroup",
       "ec2:CreateVpcPeeringConnection",
       "ec2:DeleteVpcPeeringConnection",
-      "ec2:DescribeVpcPeeringConnections",
     ]
     resources = [
       aws_vpc.assessment.arn,
--- a/terraformer_permissionsboundary.tf
+++ b/terraformer_permissionsboundary.tf
@@ -23,7 +23,7 @@ data "aws_iam_policy_document" "terraformer_permissions_boundary_policy_doc" {
     condition {
       test = "StringEquals"
       values = [
-        var.tags["Team"],
+        lookup(var.tags, "Team", "Undefined Team tag value"),
       ]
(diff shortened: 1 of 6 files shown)


variables.tf
```diff
@@ -330,6 +330,18 @@ variable "tags" {
   default     = {}
 }
 
+variable "terraformer_permissions_boundary_policy_description" {
+  type        = string
+  description = "The description to associate with the IAM permissions boundary policy attached to the Terraformer instance role in order to protect the foundational resources deployed in this account."
+  default     = "Defines the IAM permissions boundary for Terraformer instances in order to protect the foundational resources deployed in this account."
```
jsf9k: I think this language from the `description` field is clearer and can simply be repeated:
```suggestion
  default     = "The IAM permissions boundary policy attached to the Terraformer instance role in order to protect the foundational resources deployed in this account."
```

terraformer_policy.tf
```diff
@@ -1,70 +1,32 @@
...
-    ]
-    effect = "Deny"
-    resources = [
-      aws_iam_role.guacamole_instance_role.arn,
-      aws_iam_role.samba_instance_role.arn,
-      aws_iam_role.terraformer_instance_role.arn,
-    ]
-  }
 
   # Allow use of the KMS key used to encrypt COOL AMIs.
```
jsf9k: I think we should have the same comments for each of these `statement`s as are used in the corresponding `statement`s in `terraformer_permissionsboundary.tf`.  The latter are more informative, and it's the same code.

jsf9k (approved):
Very strong work here, @dav3r.  @mcdonnnj was definitely on the right path with his suggestion.

I have a couple of suggestions for your consideration, but if you disagree I still approve the PR.

terraformer_policy.tf
dav3r: Good call - updated in https://github.com/cisagov/cool-assessment-terraform/pull/215/commits/df75069c286cd745eac69f96cc66d5cd38772bb0.

jsf9k (approved):
:heart_eyes_cat:

jasonodoom (approved):

terraformer_permissionsboundary.tf
```diff
@@ -0,0 +1,291 @@
...
+  statement {
+    actions = [
+      "ec2:Describe*",
+      "ec2:Get*",
+      "ec2:List*",
+    ]
+    condition {
+      test = "StringEquals"
+      values = [
+        var.tags["Team"],
```
mcdonnnj: This variable is optional. If it is not provided, or the key is absent from the provided value, I believe this will break. I think `lookup()` (with a default value) is the only way to safely access a possibly undefined value.

terraformer_permissionsboundary.tf
```diff
@@ -0,0 +1,291 @@
...
+  # Deny modification of any resources tagged by the team that deploys this
+  # root module, but allow anything else.
+  statement {
+    actions = [
+      "*",
+    ]
+    condition {
+      test = "StringNotEquals"
+      values = [
+        var.tags["Team"],
```
mcdonnnj: Same comment about unsafe key access.

terraformer_policy.tf
```diff
@@ -1,72 +1,35 @@
...
+  # Deny modification of any resources tagged by the team that deploys this
+  # root module, but allow anything else.
   statement {
+    actions = [
+      "*",
+    ]
     condition {
       test = "StringNotEquals"
       values = [
         var.tags["Team"],
```
mcdonnnj: Guess I missed this in previous reviews but same question about unsafe access.

terraformer_policy.tf
```diff
@@ -1,72 +1,35 @@
...
-  # _are not_ tagged as being created by the team that deploys this
-  # root module, with the exception of IAM.
-  #
-  # We will attach the arn:aws:iam::aws:policy/ReadOnlyAccess policy
-  # to the same role to which the policy document will be attached, which
-  # will give it read-only access to all IAM resources.  We require
-  # IAM access to be as read-only as possible in order to stop the
-  # Terraformer instance from defeating the Terraformer policy by
-  # creating new users, policies, roles, etc.
+  # Deny modification of any resources tagged by the team that deploys this
```
mcdonnnj: I'm not a fan of this phrasing because there's no explicit `Deny` going on here. I would rather frame it terms of what is allowed since this is an allow statement.

dav3r: Please take a look at https://github.com/cisagov/cool-assessment-terraform/pull/215/commits/b7a9b68396c9557f73bcc6601cff39e4bc64f282 and let me know what you think.

terraformer_permissionsboundary.tf
dav3r: Excellent catch!  Please let me know if you're good with https://github.com/cisagov/cool-assessment-terraform/pull/215/commits/9b061a379b7d598125e40e8a711414ed270ddff5.

terraformer_permissionsboundary.tf
dav3r: See https://github.com/cisagov/cool-assessment-terraform/pull/215/commits/9b061a379b7d598125e40e8a711414ed270ddff5.

terraformer_policy.tf
dav3r: See https://github.com/cisagov/cool-assessment-terraform/pull/215/commits/9b061a379b7d598125e40e8a711414ed270ddff5.

terraformer_permissionsboundary.tf
```diff
@@ -0,0 +1,291 @@
...
+      "kms:Decrypt",
+      "kms:ReEncryptFrom",
+    ]
+    resources = [
+      data.terraform_remote_state.images.outputs.ami_kms_key.arn
+    ]
+    sid = "AllowUsingCOOLAMIKMSKey"
+  }
+
+  # Allow the launching of new instances in the operations subnet,
```
mcdonnnj: This also allows launching in the private subnet no?

terraformer_permissionsboundary.tf
```diff
@@ -0,0 +1,291 @@
...
+      aws_vpc_endpoint.s3.arn,
+    ]
+    sid = "AllowModifyingS3Endpoint"
+  }
+
+  # Don't allow Terraformer instances to touch the CloudFormation foo
+  # put in place by Control Tower.  This is not covered by the earlier
+  # statement allowing full access to resources that are not tagged as
+  # belonging to the team that deploys this root module, since CloudFormation
+  # resources do not accept tags.
```
mcdonnnj: Maybe I'm just having a time while reading this comment, but it starts by talking about what's being denied and then says we're doing it because it's not covered by an allow?

mcdonnnj (approved):
Two small questions about a couple comments but otherwise LGTM ✔

terraformer_permissionsboundary.tf
dav3r: You are correct - good catch, see https://github.com/cisagov/cool-assessment-terraform/pull/215/commits/39690032e1aff1d560d939d6cf3b7a171479e366.

terraformer_permissionsboundary.tf
dav3r: Take a look below at my attempt to make this read better:

```
  # Don't allow Terraformer instances to touch the CloudFormation foo put in
  # place by Control Tower.  CloudFormation resources do not accept tags and
  # thus cannot be tagged with a team tag, which is why this statement is needed
  # (an earlier statement prevents modification of any resources tagged by the
  # team that deploys this root module).
```

I'm open to suggestions if this doesn't work for you.

mcdonnnj: That works for me. 👍

dav3r: Cool, please see https://github.com/cisagov/cool-assessment-terraform/pull/215/commits/7adcc68974453e1d9de2f9fe7a180c723d27341e.

jasonodoom (approved):

PR #240, opened 2024-05-14
Add variable for publishing egress IP addresses

dav3r:
#  #

## 🗣 Description ##

This PR adds a new Terraform variable (`publish_egress_ip_addresses`) that specifies whether EC2 instances in the operations subnet should be tagged to indicate that their public IP addresses may be published.

This is useful for deconfliction purposes and it allows us to control publishing of IP addresses on a per-assessment (per-Terraform workspace) basis, which is a requirement of https://github.com/cisagov/cool-system-internal/issues/125.

IP address publishing will be handled via the code in [`cisagov/publish-egress-ip-lambda`](https://github.com/cisagov/publish-egress-ip-lambda) and [`cisagov/publish-egress-ip-terraform`](https://github.com/cisagov/publish-egress-ip-terraform).

## 🧪 Testing ##

I tested this by running a `terraform plan` for an environment with:
* `publish_egress_ip_addresses = true`
* No value set for `publish_egress_ip_addresses`

In the first case, I confirmed that the plan wanted to set `"Publish Egress" = "true"` for all of the correct instances/EIPs.
In the second case, I confirmed that the plan wanted to set `"Publish Egress" = "false"` for all of the correct instances/EIPs.

## ✅ Post-merge checklist ##

- [x] Coordinate with the COOL Operations team to ensure that this new variable is used correctly.
- [x] Update all Staging and Production tfvars files to include `publish_egress_ip_addresses`


jsf9k (approved):

dv4harr10 (approved):

debiandesktop_ec2.tf
```diff
@@ -56,7 +56,8 @@ resource "aws_instance" "debiandesktop" {
   }
   subnet_id = aws_subnet.operations.id
   tags = {
-    Name = format("DebianDesktop%d", count.index)
+    Name             = format("DebianDesktop%d", count.index)
+    "Publish Egress" = var.publish_egress_ip_addresses
```
mcdonnnj, line 60: Will this value massage to the `True` or `False` that was in use before? I'm concerned this will interpolate to a string version of `true` or `false` which doesn't quite match up.

mcdonnnj (commented):
This seems pretty solid. In addition to my one feedback item I would like to raise maybe making an issue to expand functionality to allow control over which instances of a given type have their IP information published e.g. I want `Windows0` to be private but `Windows1` to be published.

debiandesktop_ec2.tf
dav3r: You are good!  After this PR is merged, I was planning to modify the lambda code [here](https://github.com/cisagov/publish-egress-ip-lambda/blob/bfb38499afdb29ac622f687a364088e8bf19e119/src/lambda_handler.py#L97-L110) to account for this change from `True` to `true`.

I already had it written, but haven't committed it yet:

Original:
```python
        if tags.get(publish_egress_tag_name, str(False)) != str(True):
```
New:
```python
        if tags.get(publish_egress_tag_name, "false").lower() != "true":
```

pentestportal_ec2.tf
```diff
@@ -80,7 +80,8 @@ resource "aws_eip" "pentestportal" {
   provider = aws.provisionassessment
 
   tags = {
-    Name             = format("PentestPortal%d EIP", count.index + 1)
+    Name = format("PentestPortal%d EIP", count.index + 1)
+    # This instance type is always excluded from egress IP publishing
     "Publish Egress" = "False"
```
mcdonnnj: Based on https://github.com/cisagov/cool-assessment-terraform/pull/240#discussion_r1600570216 I think this should be changed.

```suggestion
    "Publish Egress" = "false"
```

dav3r: Anything other than `true` won't actually matter, but I still agree with you. 😄

dav3r:
> This seems pretty solid. In addition to my one feedback item I would like to raise maybe making an issue to expand functionality to allow control over which instances of a given type have their IP information published e.g. I want `Windows0` to be private but `Windows1` to be published.

@mcdonnnj That's a good idea and we should keep it in mind, but until someone comes to us with that as a requirement, I don't think we need to worry too much about it.

Do I have your approval on this PR?

mcdonnnj:
> > This seems pretty solid. In addition to my one feedback item I would like to raise maybe making an issue to expand functionality to allow control over which instances of a given type have their IP information published e.g. I want `Windows0` to be private but `Windows1` to be published.
> 
> @mcdonnnj That's a good idea and we should keep it in mind, but until someone comes to us with that as a requirement, I don't think we need to worry too much about it.
> 
> Do I have your approval on this PR?

Yeah I meant it as a "feature idea" issue just to document the thought.

mcdonnnj (approved):
LGTM ✔

dav3r:
> > > This seems pretty solid. In addition to my one feedback item I would like to raise maybe making an issue to expand functionality to allow control over which instances of a given type have their IP information published e.g. I want `Windows0` to be private but `Windows1` to be published.
> > 
> > 
> > @mcdonnnj That's a good idea and we should keep it in mind, but until someone comes to us with that as a requirement, I don't think we need to worry too much about it.
> > Do I have your approval on this PR?
> 
> Yeah I meant it as a "feature idea" issue just to document the thought.

Understood - I put this idea in https://github.com/cisagov/cool-assessment-terraform/issues/242.  Feel free to make any edits as you see fit.

PR #255, opened 2024-10-23
Make the `sed` command work in MacOS

jsf9k:
## 🗣 Description ##

This pull request modifies the `sed` command in the script for tainting instances to work on both MacOS and Linux.  It also fixes a bug where the Samba instance was not being ignored as it should have been.

- MacOS's version of `sed` doesn't support long options, so we have to use short options in our script for tainting instances.
- MacOS's version of `sed` doesn't use extended regexes by default, so we have to add the `-E` flag.
- The Samba instances should not be tainted since the assessors do not have access to those instances.

## 🧪 Testing ##

All automated tests pass.  I tested this change locally and verified that it functioned on Linux, and @coffeegist did the same on MacOS.

### DIFF
--- a/taint_all_assessment_instances.sh
+++ b/taint_all_assessment_instances.sh
@@ -17,10 +17,12 @@ export AWS_PROFILE
 export AWS_SHARED_CREDENTIALS_FILE
 export AWS_DEFAULT_REGION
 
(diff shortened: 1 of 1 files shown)


dv4harr10 (approved):

dav3r (approved):
👍

PR #273, opened 2025-08-25
Fix error in `cloud-init` script

jsf9k:
## 🗣 Description ##

This pull request fixes a `cloud-init` script that recently broke.

The script wasn't working as it stood, so it had to be fixed.  @m1j09830 noticed this and brought it to my attention.

The problem was that `nvme id-ctrl -v` no longer shows the non-NVMe device name.  The non-NVMe device name shows up in Linux as a symlink to the appropriate NVMe device, though, so none of the chicanery where we figure out the NVMe device name is actually necessary.  (I'm not sure it ever was.)  We can use the non-NVMe device name directly.

## 🧪 Testing ##

I verified that the GoPhish instance types deploys successfully with this change in both our `dev-a` and `production` COOL deployments.

- [x] Bump major, minor, patch, pre-release, and/or build versions [as appropriate](https://semver.org/#semantic-versioning-specification-semver) via the `bump_version` script *if* this repository is versioned *and* the changes in this PR [warrant a version bump](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api).

## ✅ Pre-merge checklist ##

- [x] Finalize version.

## ✅ Post-merge checklist ##

- [x] Create a release (necessary if and only if the version was bumped).


dav3r (approved):
Excellent, love that it's simpler now. 👍

cloud-init/ebs-disk-setup.tpl.sh
```diff
@@ -30,44 +30,19 @@ while [ "$(lsblk | grep --count ' disk')" -lt "${num_disks}" ]; do
...
-  # looking for.
-  set +o errexit
-  set +o pipefail
-  non_nvme_device_name=$(nvme id-ctrl -v "$nvme_device" | grep -o "${device_name}")
-  set -o errexit
-  set -o pipefail
-
-  if [ "$non_nvme_device_name" = "${device_name}" ]; then
-    # We've found our device
+# Create a file system on the EBS volume if one was not already there.
```
mcdonnnj: Since we don't tend to use periods on single line comments (now that this is no longer a multi-line comment).

```suggestion
# Create a file system on the EBS volume if one was not already there
```

cloud-init/ebs-disk-setup.tpl.sh
```diff
@@ -30,44 +30,19 @@ while [ "$(lsblk | grep --count ' disk')" -lt "${num_disks}" ]; do
...
-    mount UUID="$uuid" --options "${mount_options}" "${mount_point}"
-
-    # Save the mount point in fstab, so the file system is
-    # remounted if the instance is rebooted
-    echo "# ${label}" >> /etc/fstab
-    echo "UUID=$uuid ${mount_point} ${fs_type} ${mount_options} 0 2" >> /etc/fstab
-  fi
-done
+# Save the mount point in fstab, so the file system is remounted if
+# the instance is rebooted
```
mcdonnnj: Since we tend to use periods on multi-line comments.

```suggestion
# the instance is rebooted.
```

mcdonnnj (approved):
LGTM ✔

PR #295, opened 2026-07-07
Add a loop to retry setting of Nessus port until it succeeds or exhausts the allowed number of retries

jsf9k:
## 🗣 Description ##

This pull request adds a loop to `cloud-init/nessus-setup.sh` to retry the setting of the Nessus port until it either succeeds or exhausts the allowed number of retries.

I have found that sometimes the setting of the Nessus port fails because "global.db is not ready yet", but it eventually succeeds after a few minutes if retried.

## 🧪 Testing ##

I successfully deployed a new Nessus instance with these changes and verified that they function as expected.  No manual re-running of cloud-init scripts was necessary, which is a great improvement.

I also temporarily set the maximum number of retries to one and verified that the logic stops there even if the attempt to set the Nessus web port fails.

- [x] Bump major, minor, patch, pre-release, and/or build versions [as appropriate](https://semver.org/#semantic-versioning-specification-semver) via the `bump_version` script *if* this repository is versioned *and* the changes in this PR [warrant a version bump](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api).

## ✅ Pre-merge checklist ##

- [x] Finalize version.

## ✅ Post-merge checklist ##

- [x] Create a release (necessary if and only if the version was bumped).


cloud-init/nessus-setup.sh
```diff
@@ -77,16 +77,20 @@ else
...
-fi
+while true; do
+  # shellcheck disable=SC2154
+  echo "Attempting to set Nessus web server port to ${nessus_web_server_port}..."
+  $nessus_sbin_path/nessuscli fix --set xmlrpc_listen_port="${nessus_web_server_port}"
+  xmlrpc_listen_port_rc="$?"
+
+  # No need to echo anything if successful since nessuscli will do it for us
+  if [ "$xmlrpc_listen_port_rc" -ne 0 ]; then
+    echo "ERROR: Setting Nessus web server port was unsuccessful"
```
mcdonnnj: I think it would be beneficial to add a message mentioning that it will be retried.

mcdonnnj (commented):
Unless you have other changes on the way did you have a reason not to bump the version?

cloud-init/nessus-setup.sh
```diff
@@ -77,16 +77,20 @@ else
...
+    sleep 60
+  else
+    break
+  fi
+done
```
dav3r: I'd prefer to see a loop that will timeout after a certain number of tries, rather than a never-ending one.

dav3r (changes requested):
Please take a look at my comment and let me know what you think.

cloud-init/nessus-setup.sh
jsf9k: Please see commit [f379a00](https://github.com/cisagov/cool-assessment-terraform/pull/295/commits/f379a008dc15c614108d4f8453a507c156f20529).

cloud-init/nessus-setup.sh
jsf9k: Please see commit [5d989d4](https://github.com/cisagov/cool-assessment-terraform/pull/295/commits/5d989d43ef9e856b44c96176fbedacf47c59c2e7).

jsf9k:
I re-tested these changes twice after commit [5d989d4](https://github.com/cisagov/cool-assessment-terraform/pull/295/commits/5d989d43ef9e856b44c96176fbedacf47c59c2e7).  I verified:
- That the code works as-is
- That the maximum number of attempts is not exceeded by temporarily setting it equal to one

jsf9k:
> Unless you have other changes on the way did you have a reason not to bump the version?

Please see commit [c804f78](https://github.com/cisagov/cool-assessment-terraform/pull/295/commits/c804f78fc2a9d948fcf0ff83af1cd09f1b6c9fdc).

mcdonnnj (approved):

cloud-init/nessus-setup.sh
dav3r: That looks great, thanks. 👍

dav3r (approved):
🎡 👍

PR #299, opened 2026-08-12
Use fewer ACL rules for dynamic inbound ports in the operations subnet

### LINKED ISSUE #288: Inefficient use of NACL rules for `inbound_ports_allowed`
# 🐛 Summary #

Adding a lot of `inbound_ports_allowed` can result in hitting the AWS hard limit for network ACL rules (currently 40) during a `terraform apply`.

## To reproduce ##

1. Set up an environment with inbound ports allowed similar to [this](https://github.com/cisagov/cool-assessment-tfvars/blob/8f34dac9a9fab5e0b0918b64f1920fcc85daad3a/production/env82-production.tfvars).
1. Attempt to apply the terraform.

## Expected behavior ##

The `terraform apply` should complete without error and the security groups and network ACL rules should be created successfully in AWS.

## Any helpful log output or screenshots ##

If the limit for NACL rules is exceeded, Terraform gives one or more errors like this:

```console
Error: creating EC2 Network ACL (acl-01234567890abcdef) Rule (egress: false)(156): operation error EC2: CreateNetworkAclEntry, https response error StatusCode: 400, RequestID: 0b8293e6-d21a-428a-97bd-413e79c7603f, api error NetworkAclEntryLimitExceeded: The maximum number of network acl entries has been reached.
│ 
│   with aws_network_acl_rule.operations_ingress_from_anywhere_via_allowed_ports["tcp_123_123"],
│   on [operations_acl_rules.tf](http://operations_acl_rules.tf/) line 122, in resource "aws_network_acl_rule" "operations_ingress_from_anywhere_via_allowed_ports":
│  122: resource "aws_network_acl_rule" "operations_ingress_from_anywhere_via_allowed_ports" {
│
```

We should find a way to minimize the necessary number of NACL rules such that we will not run into this hard limit.  For example, one naive way to do this would be to simply create one rule per protocol/source/allow/deny containing the range from the lowest to the highest ports required.  Since the security groups provide granular security, the NACLs could be more coarsely defined.

jsf9k:
## 🗣 Description ##

This pull request modifies the code to compute maximal ranges for inbound ports in the operations subnet that are opened per the `*.tfvars` file.

Resolves #288.

AWS only allows 40 ACL rules per subnet.  With cisagov/egress-assess-packer#67 requiring more inbound ports be opened up for that instance type we have hit that limit.

## 🧪 Testing ##

I tested these changes by successfully applying them to `env38` in our `production` COOL deployment.  I looked at the ACL rules created for the operations subnet and they look correct.


dav3r:
This is output from an AI review:

## Summary of the change

This PR reduces the number of network ACL rules consumed by `aws_network_acl_rule.operations_ingress_from_anywhere_via_allowed_ports`. Previously it created one ACL rule per distinct `{protocol, from_port, to_port}` entry across all instantiated operations instances. Now it collapses those down to **one rule per protocol**, spanning from the minimum `from_port` to the maximum `to_port` seen for that protocol. This works around the AWS limit of ~40 rules per ACL. The mechanism is sound and the Terraform is correct.

## Correctness — looks good

- The new `inbound_ports_allowed_for_acl` map exposes `index`, `from_port`, `protocol`, `to_port`, which exactly matches what the ACL resource references (`each.value.index`, etc.), so the `for_each` swap is consistent.
- `index` was dropped from `union_of_inbound_ports_allowed`, and nothing else consumes it — the per-instance SG files (`kali_sg.tf`, etc.) read `var.inbound_ports_allowed` directly, not this local. So no dangling references.
- The empty case is safe: with no instantiated instances, `union_of_inbound_ports_allowed_protocols` is `[]`, all downstream maps are empty, and `min([]...)`/`max([]...)` are never invoked (the comprehension produces no elements). No apply-time error.
- `min([...]...)` / `max([...]...)` with the `...` spread is valid Terraform.
- Rule numbering is fine: protocol count is tiny (tcp/udp), so `150 + index` stays at 150/151 and won't collide with the next rule block at 170.

## The thing I'd flag before approving

**The consolidation meaningfully broadens the ACL, and can expose deliberately-excluded ports at the NACL layer.**

Because you now open a single contiguous `min→max` range per protocol, every port *between* the explicitly-requested ports is also opened at the network ACL. Example: if an instance requests `tcp 443` and `tcp 9000–9009`, the ACL now allows `tcp 443–9009`, i.e. 444–8999 are open at the subnet boundary where they weren't before.

The PR description correctly notes that the **security groups still restrict actual traffic**, so effective exposure is unchanged — that's a reasonable defense-in-depth trade to stay under the rule limit. But there's a subtler interaction worth checking:

The other ACL rules in `operations_acl_rules.tf` deliberately carve *around* two sensitive ports — RDP `3389` and Cobalt Strike teamserver `50050` — by splitting the ephemeral ranges (`1024–3388`, `3390–50049`, `50051–65535`). This new consolidated allow rule sits at rule number **150/151**, which is evaluated *before* those carve-outs (170/180/200), and there is no explicit `deny` for 3389 or 50050. So **if the min→max range for a protocol happens to span 3389 or 50050, this rule will open them at the NACL** — silently undoing the exclusion those other rules were written to preserve. Only the security groups would then stand between the internet and those ports.

Whether this actually triggers depends on the real `inbound_ports_allowed` values in the deployment tfvars. It's worth confirming that no protocol's consolidated range straddles 3389 or 50050 — and if that's an acceptable risk given SG enforcement, consider adding a comment near `inbound_ports_allowed_for_acl` documenting that the range is intentionally coarse and relies on the SGs, so a future reader doesn't assume the ACL is precise.

## Minor / optional

- Pre-existing (not introduced here): if anyone ever puts an `icmp` entry in `inbound_ports_allowed`, the `from_port`/`to_port` min/max math is semantically meaningless for ICMP and the ACL rule wouldn't set `icmp_type`/`icmp_code`. The old code had the same issue, so it's not a regression — just worth knowing the new range math makes it slightly more surprising.
- The testing note (applied to `env38`, ACL rules look correct) is good. For extra confidence, it'd be worth explicitly verifying in that environment that 3389 and 50050 are *not* reachable at the NACL after the change.

Overall: functionally correct and achieves its stated goal. The one real ask is to confirm the 3389/50050 exposure question and, if it's fine, document the intentional coarsening. Everything else is clean.

dav3r:
AI suggestions:

Replying to my earlier review (https://github.com/cisagov/cool-assessment-terraform/pull/299#issuecomment-5270056885) regarding protecting the sensitive ports (3389 RDP, 50050 Cobalt Strike teamserver) from the broadened `min→max` range.

The cleanest fix leverages how AWS network ACLs evaluate rules: they're processed in ascending `rule_number` order and **the first match wins**. So put explicit `deny` rules for the sensitive ports at a rule number *below* the broad allow (150/151), and they'll shadow it no matter what range the consolidation produces.

## Option A (recommended): explicit deny rules ahead of the broad allow

Data-driven, costs only 1–2 rules, and is robust to whatever `min→max` range gets computed.

In `locals.tf`:

```hcl
# Ports that must never be publicly reachable at the operations subnet
# ACL, regardless of the ranges produced by inbound_ports_allowed.
# These deny rules are numbered below the "allowed ports" rule
# (150/151) so they take precedence (first-match wins in a NACL).
operations_denied_public_ports = {
  rdp        = { port = 3389, rule_number = 145 }
  teamserver = { port = 50050, rule_number = 146 }
}
```

In `operations_acl_rules.tf` (place it just before the `operations_ingress_from_anywhere_via_allowed_ports` rule):

```hcl
# Explicitly deny public access to sensitive ports (RDP, Cobalt Strike
# teamserver) so the broadened "allowed ports" range below cannot
# inadvertently expose them at the NACL layer.
resource "aws_network_acl_rule" "operations_deny_sensitive_public_ports" {
  provider = aws.provisionassessment
  for_each = local.operations_denied_public_ports

  cidr_block     = "0.0.0.0/0"
  egress         = false
  from_port      = each.value.port
  network_acl_id = aws_network_acl.operations.id
  protocol       = "tcp"
  rule_action    = "deny"
  rule_number    = each.value.rule_number
  to_port        = each.value.port
}
```

Why this works well here:
- Both 3389 (RDP) and 50050 (teamserver) are TCP, so two rules cover them.
- Neither port appears in the private-subnet allow rules (100–140 are ssh/winrm/vnc/http/https), so a blanket `0.0.0.0/0` deny won't break Ansible/Terraformer traffic — Ansible reaches Windows via WinRM/5986, not RDP.
- Future-proof: to protect another port later, add one line to the map.

**Caveat to confirm:** if any legitimate flow *from the private subnet* ever needs 3389 or 50050, a `0.0.0.0/0` deny at 145/146 would block it too (NACLs don't do longest-prefix; the lower rule number just wins). If that's a concern, add a private-subnet `allow` for those ports at an even lower number (e.g. 141–144) so private traffic is permitted before the public deny. Based on the current design I don't think it's needed, but worth a moment's thought.

## Option B: compute ranges that exclude the sensitive ports

Instead of a single `min→max` per protocol, split the range around 3389/50050 so the allow rules never cover them. This keeps the ACL precise, but reintroduces rule-count growth (potentially 2–3 sub-ranges per protocol), which fights the purpose of the PR, and makes the locals logic more complex. Only worth it if there's a hard requirement that the NACL never allow *any* unrequested port, not just the two sensitive ones.

## Option C: accept the broadening, document it, rely on SGs

Essentially the current PR state. Effective exposure is unchanged because the security groups still filter, so it's defensible. The risk is purely defense-in-depth erosion and the surprise factor for a future reader. If you go this route, at least add a comment near `inbound_ports_allowed_for_acl` explaining the range is intentionally coarse.

## Recommendation

Option A. It restores the deliberate 3389/50050 exclusion the original ACL design had, keeps you well under the 40-rule limit (two rules), and is self-documenting via the `locals` map. It complements the PR rather than fighting it.

jsf9k:
> AI suggestions:
> 
> Replying to my earlier review ([#299 (comment)](https://github.com/cisagov/cool-assessment-terraform/pull/299#issuecomment-5270056885)) regarding protecting the sensitive ports (3389 RDP, 50050 Cobalt Strike teamserver) from the broadened `min→max` range.
> 
> The cleanest fix leverages how AWS network ACLs evaluate rules: they're processed in ascending `rule_number` order and **the first match wins**. So put explicit `deny` rules for the sensitive ports at a rule number _below_ the broad allow (150/151), and they'll shadow it no matter what range the consolidation produces.
> 
> ## Option A (recommended): explicit deny rules ahead of the broad allow
> Data-driven, costs only 1–2 rules, and is robust to whatever `min→max` range gets computed.
> 
> In `locals.tf`:
> 
> ```hcl
> # Ports that must never be publicly reachable at the operations subnet
> # ACL, regardless of the ranges produced by inbound_ports_allowed.
> # These deny rules are numbered below the "allowed ports" rule
> # (150/151) so they take precedence (first-match wins in a NACL).
> operations_denied_public_ports = {
>   rdp        = { port = 3389, rule_number = 145 }
>   teamserver = { port = 50050, rule_number = 146 }
> }
> ```
> 
> In `operations_acl_rules.tf` (place it just before the `operations_ingress_from_anywhere_via_allowed_ports` rule):
> 
> ```hcl
> # Explicitly deny public access to sensitive ports (RDP, Cobalt Strike
> # teamserver) so the broadened "allowed ports" range below cannot
> # inadvertently expose them at the NACL layer.
> resource "aws_network_acl_rule" "operations_deny_sensitive_public_ports" {
>   provider = aws.provisionassessment
>   for_each = local.operations_denied_public_ports
> 
>   cidr_block     = "0.0.0.0/0"
>   egress         = false
>   from_port      = each.value.port
>   network_acl_id = aws_network_acl.operations.id
>   protocol       = "tcp"
>   rule_action    = "deny"
>   rule_number    = each.value.rule_number
>   to_port        = each.value.port
> }
> ```
> 
> Why this works well here:
> 
> * Both 3389 (RDP) and 50050 (teamserver) are TCP, so two rules cover them.
> * Neither port appears in the private-subnet allow rules (100–140 are ssh/winrm/vnc/http/https), so a blanket `0.0.0.0/0` deny won't break Ansible/Terraformer traffic — Ansible reaches Windows via WinRM/5986, not RDP.
> * Future-proof: to protect another port later, add one line to the map.
> 
> **Caveat to confirm:** if any legitimate flow _from the private subnet_ ever needs 3389 or 50050, a `0.0.0.0/0` deny at 145/146 would block it too (NACLs don't do longest-prefix; the lower rule number just wins). If that's a concern, add a private-subnet `allow` for those ports at an even lower number (e.g. 141–144) so private traffic is permitted before the public deny. Based on the current design I don't think it's needed, but worth a moment's thought.
> 
> ## Option B: compute ranges that exclude the sensitive ports
> Instead of a single `min→max` per protocol, split the range around 3389/50050 so the allow rules never cover them. This keeps the ACL precise, but reintroduces rule-count growth (potentially 2–3 sub-ranges per protocol), which fights the purpose of the PR, and makes the locals logic more complex. Only worth it if there's a hard requirement that the NACL never allow _any_ unrequested port, not just the two sensitive ones.
> 
> ## Option C: accept the broadening, document it, rely on SGs
> Essentially the current PR state. Effective exposure is unchanged because the security groups still filter, so it's defensible. The risk is purely defense-in-depth erosion and the surprise factor for a future reader. If you go this route, at least add a comment near `inbound_ports_allowed_for_acl` explaining the range is intentionally coarse.
> 
> ## Recommendation
> Option A. It restores the deliberate 3389/50050 exclusion the original ACL design had, keeps you well under the 40-rule limit (two rules), and is self-documenting via the `locals` map. It complements the PR rather than fighting it.

Please see commit [d0411f2](https://github.com/cisagov/cool-assessment-terraform/pull/299/commits/d0411f270b87bc29f7f97c2a843cba46a6cc0402).

dav3r (approved):
Strong work, looks great, excelsior! 🚀

dav3r (approved):

